PR #24, opened 2022-11-16
add changelog.md

akanz1:
When getting an extension update it's nice to take a quick look at new features and changes that come with it.

This PR adds a CHANGELOG.md which should be picked up by the vscode extension to display a changelog tab.

I've added the most recent releases and ideally, this would also include all changes from ruff between the different versions of the vscode extension. 

However, lowest maintenance effort is probably to simply copy the auto-generated vscode-ruff release notes into the CHANGELOG.md or to just point a link to the github release notes (see example below), which is not nice but at least users can access the release notes with a click.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,24 @@
+# Changelog
+
+See [here](https://github.com/charliermarsh/ruff/releases) for the Ruff release notes
+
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thanks!

PR #26, opened 2022-11-20
Show check summary when hovering on checkcode in noqa comment

### LINKED ISSUE #25: [Feature idea] Show check summary when the cursor hovers on a `noqa: comment`
image

It's difficult to infer the check summary from the check code (e.g. `C402`). I think it's useful if the ruff extension shows a tooltip describing the check.

harupy:
Resolve #25.

Demo:

https://user-images.githubusercontent.com/17039389/202880358-a220f900-5ad5-4aa2-9099-bb081c90c3e6.mov


bundled/tool/server.py
```diff
@@ -223,6 +252,101 @@ def _get_settings_by_document(document: workspace.Document | None):
...
+        # If we're loading from the bundle, use the absolute path.
+        argv = [
+            os.fspath(pathlib.Path(__file__).parent.parent / "libs" / "bin" / TOOL_MODULE),
+        ]
+    else:
+        # If the interpreter is same as the interpreter running this process then run
+        # as module.
+        argv = [os.path.join(sysconfig.get_path("scripts"), TOOL_MODULE)]
+
+    argv += ["--explain", code, "."]
```
harupy, line 299: We need something that can map a check code to its summary.

harupy: Implementation I used for the demo

```diff
diff --git a/src/cli.rs b/src/cli.rs
index b1cd519..ed9c7ae 100644
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -109,6 +109,9 @@ pub struct Cli {
     /// The name of the file when passing it through stdin.
     #[arg(long)]
     pub stdin_filename: Option<String>,
+    /// Explain a check.
+    #[arg(long)]
+    pub explain: Option<String>,
 }
 
 impl Cli {
diff --git a/src/main.rs b/src/main.rs
index 5c5e605..d32bfaa 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,7 @@
 use std::io::{self, Read};
 use std::path::{Path, PathBuf};
 use std::process::ExitCode;
+use std::str::FromStr;
 use std::sync::mpsc::channel;
 use std::time::Instant;
 
@@ -284,6 +285,18 @@ fn inner_main() -> Result<ExitCode> {
         show_settings(configuration, project_root, pyproject);
         return Ok(ExitCode::SUCCESS);
     }
+    if let Some(code) = cli.explain {
+        match CheckCode::from_str(&code) {
+            Ok(code) => {
+                println!("{}", code.kind().summary());
+                return Ok(ExitCode::SUCCESS);
+            }
+            Err(_) => {
+                eprintln!("Unknown check code");
+                return Ok(ExitCode::FAILURE);
+            }
+        }
+    }
 
     // Extract settings for internal use.
     let autofix = configuration.fix;
```

bundled/tool/server.py
```diff
@@ -84,6 +85,35 @@ def did_save(params: lsp.DidSaveTextDocumentParams) -> None:
     diagnostics: list[lsp.Diagnostic] = _linting_helper(document)
     LSP_SERVER.publish_diagnostics(document.uri, diagnostics)
 
+NOQA_REGEX = re.compile(r"(?i:# noqa)(?::\s?(?P<codes>([A-Z]+[0-9]+(?:[,\s]+)?)+))?")
+CODE_REGEX = re.compile(r"[A-Z]+[0-9]+")
+
+
+@LSP_SERVER.feature(lsp.HOVER)
+def hover(params: lsp.HoverParams) -> None:
```
harupy, line 93: Another option to show a check summary is via Diagnostics.


https://user-images.githubusercontent.com/17039389/202880911-fafb038d-9ac5-45dc-b790-8946d40a52cc.mov

In this demo, I tweaked the ruff source code to show check code locations as violations.

bundled/tool/server.py
harupy: Instead, we could provide a CLI option that lists checks ruff supports:

```
ruff --list-errors ...

C401: ...
C402: ...
```

and then parse its output to create a mapping to convert a check code to its summary when the LSP is launched.

harupy:
```diff
diff --git a/src/cli.rs b/src/cli.rs
index b1cd519..764f562 100644
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -109,6 +109,9 @@ pub struct Cli {
     /// The name of the file when passing it through stdin.
     #[arg(long)]
     pub stdin_filename: Option<String>,
+    /// Show all available checks
+    #[arg(long)]
+    pub show_checks: bool,
 }
 
 impl Cli {
diff --git a/src/main.rs b/src/main.rs
index 5c5e605..c0d6d0e 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,6 +26,7 @@ use log::{debug, error};
 use notify::{raw_watcher, RecursiveMode, Watcher};
 #[cfg(not(target_family = "wasm"))]
 use rayon::prelude::*;
+use strum::IntoEnumIterator;
 use walkdir::DirEntry;
 
 /// Shim that calls par_iter except for wasm because there's no wasm support in
@@ -284,6 +285,13 @@ fn inner_main() -> Result<ExitCode> {
         show_settings(configuration, project_root, pyproject);
         return Ok(ExitCode::SUCCESS);
     }
+    if cli.show_checks {
+        // Show check summaries.
+        for c in CheckCode::iter() {
+            println!("{}: {}", c.kind().code().as_ref(), c.kind().summary());
+        }
+        return Ok(ExitCode::SUCCESS);
+    }
 
     // Extract settings for internal use.
     let autofix = configuration.fix;
```

bundled/tool/server.py
```diff
@@ -223,6 +257,108 @@ def _get_settings_by_document(document: workspace.Document | None):
...
+        # If we're loading from the bundle, use the absolute path.
+        argv = [
+            os.fspath(pathlib.Path(__file__).parent.parent / "libs" / "bin" / TOOL_MODULE),
+        ]
+    else:
+        # If the interpreter is same as the interpreter running this process then run
+        # as module.
+        argv = [os.path.join(sysconfig.get_path("scripts"), TOOL_MODULE)]
+
+    argv += ["--show-rules", "--format", "json", "."]
```
harupy, line 311: @charliermarsh What do you think about this CLI option?

charliermarsh: Maybe more generally useful would be to add `ruff --explain F841`, similar to `clippy --explain`. Then it could print an explanation, or format it as JSON, and we could call out to that command when the user hovers over the check (rather than building up the map in advance). Would that work?

harupy: > Maybe more generally useful would be to add ruff --explain F841

Agreed. Yes, this also would work.

charliermarsh:
Sincere apologies for letting this one sit. I actually would be interested in supporting this behavior. Can I revive it?

PR #54, opened 2022-12-06
Fix 'Exception ignored in atexit callback' in jsonrpc

### LINKED ISSUE #53: atexit.register decorator causes exception
I noticed that [this line](https://github.com/charliermarsh/vscode-ruff/blob/3e5ec84d5331b11d44eebf523bc87c2a4a49358b/bundled/tool/jsonrpc.py#L136) causes an exception in the "Ruff" Output window:

Screenshot 2022-12-06 12 38 21

```
Exception ignored in atexit callback: <function ProcessManager.stop_all_processes at 0x10298a320>
TypeError: ProcessManager.stop_all_processes() missing 1 required positional argument: 'self'
```

I think this can be fixed by removing the decorator, and adding:

```
        atexit.register(self.stop_all_processes)
```

to the end of `__init__` instead. Wasn't sure if this was worthy of a PR, so opened an issue for your consideration.

eddyg:
This cleans up Ruff's "Output" tab in VSCode and fixes #53

### DIFF
--- a/bundled/tool/jsonrpc.py
+++ b/bundled/tool/jsonrpc.py
@@ -132,8 +132,8 @@ class ProcessManager:
         self._rpc: dict[str, JsonRpc] = {}
         self._lock = threading.Lock()
         self._thread_pool = ThreadPoolExecutor(10)
+        atexit.register(self.stop_all_processes)
 
-    @atexit.register  # type: ignore
     def stop_all_processes(self):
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thanks!

eddyg:
Thank _YOU_ for all the work you're putting in to Ruff! 🥇

PR #66, opened 2022-12-20
Add docs on replacing Black et al

### LINKED ISSUE #65: Adding ruff as supported linter in `python.formatting.provider`
Hey!

I've been using the extension quite a lot lately, I've been wondering if there's a way to add the ruff linter as an available option inside the `python.formatting.provider` list. 

I can handle it if needed. I think it'd be a geat addition to integrate the linter inside vscode `ms-python` extension. Instead of having `black` as default formatter and ruff doing another work right after ^^.

NB: I don't know if this is the right repo for this ^^

charliermarsh:
Resolves #65.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,21 @@ You can configure Ruff to autofix violations on-save by enabling the `source.fix
 }
 ```
 
+If you're using the VS Code Python extension, you can configure Ruff to run in lieu of another
+formatter by marking the Ruff extension as your default Python formatter:
+
+```json
+{
+    "[python]": {
+        "editor.defaultFormatter": "charliermarsh.ruff",
+        "editor.formatOnSave": true,
+        "editor.codeActionsOnSave": {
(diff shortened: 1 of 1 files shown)


max-sixty:
Am I right in thinking that with these configs `black` won't run? 

Ideally I was hoping to use `ruff` without losing `black` formatting — IIUC ruff replaces `isort` & `flake8` but not black; is that right?

charliermarsh:
Yeah, you should definitely be able to use both. This example was _specifically_ about turning Black off.

To use Ruff _and_ Black, and have them both on save, you'd want this (I think?) in your `settings.json`:

```json
{
    "[python]": {
        "editor.formatOnSave": true,
        "editor.codeActionsOnSave": {
            "source.fixAll": true
        },
    },
    "python.formatting.provider": "black"
}
```

charliermarsh:
Let me know if that works for you, or doesn't :)

PR #89, opened 2022-12-27
Confusing error message if old ruff-lsp package installed

FrancescElies:
We have some colleagues using vscode and got an error message `AttributeError: module 'ruff_lsp.server' has no attribute 'set_bundle'`, this pr gives a mo clearer error message.

When you execute `❯ .venv\Scripts\python.exe bundled\tool\server.py` with ruff-lsp v0.0.5 you get

```
  File "src\ruff-vscode\bundled\tool\server.py", line 29, in <module>
    server.set_bundle(os.fspath(BUNDLE_DIR / "libs" / "bin" / server.TOOL_MODULE))
AttributeError: module 'ruff_lsp.server' has no attribute 'set_bundle'
```

Now you get
```
  File "src\ruff-vscode\bundled\tool\server.py", line 27, in <module>
    raise RuntimeError("ruff-vscode needs at least ruff-lsp v0.0.6")
RuntimeError: ruff-vscode needs at least ruff-lsp v0.0.6
```

### DIFF
--- a/bundled/tool/server.py
+++ b/bundled/tool/server.py
@@ -23,5 +23,8 @@ update_sys_path(os.fspath(BUNDLE_DIR / "libs"))
 if __name__ == "__main__":
     from ruff_lsp import server
 
+    if not hasattr(server, "set_bundle"):
+        raise RuntimeError("ruff-vscode needs at least ruff-lsp v0.0.6")
+
     server.set_bundle(os.fspath(BUNDLE_DIR / "libs" / "bin" / server.TOOL_MODULE))
     server.start()


charliermarsh:
Thank you! This _should_ be impossible now that we give the bundled libraries higher precedence (so this should always be using the bundled version of `ruff-lsp`). I think `2022.0.25` was the only version where that could've been an error... but if not, then I'm misunderstanding something about how the bundling works 😅 Any chance you know if they were on `2022.0.25`?

FrancescElies:
Thank you!

I don't use vscode myself, we just happen to have colleagues that use it and got `AttributeError` with `2022.0.25` and got confused by that, with the latest version I tried to downgrade ruff-lsp in the currently activated venv and the vscode extension did not throw any errors, I guess the bundling worked 🙂

PR #144, opened 2023-03-01
Fix docs to refer to `fromEnvironment` as default import strategy

sjdemartini:
Closes https://github.com/charliermarsh/ruff-vscode/issues/141

The docs claimed that `useBundled` was the default import strategy, but it seems that's in fact not the case, based on
https://github.com/charliermarsh/ruff-vscode/blob/f57ae4f1cad2e7ea3a9caa307898fdfacbd180f1/src/common/settings.ts#L51

Per discussion here
https://github.com/charliermarsh/ruff-vscode/issues/141#issuecomment-1450535139, it was agreed to that `fromEnvironment` was the right default, so this updates docs to reflect that existing default.


charliermarsh:
Ah thanks! Looks like you're right.

PR #171, opened 2023-04-01
Use new Python Environment API

monosans:
The Python Environment API currently used by the extension has been deprecated.
[The base template also uses the new API.](https://github.com/microsoft/vscode-python-tools-extension-template/commit/620e863b5f6de9db092be3438c806154a7d5ae5a)


charliermarsh:
Thanks! This looks great.

PR #192, opened 2023-05-16
Better devevelopment experience

konstin:
Internal improvements that making working on the extension easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,3 +154,12 @@ This extension is based on the [Template for VS Code Python tools extensions](ht
 - To automatically format the codebase, run: `just fmt`.
 - To run lint and type checks, run: `just check`.
 - To run tests, run: `just test`.
+
+To run the extension, navigate to `src/extension.ts` and run (`F5`). You should see the LSP output
+and Python log messages in the debug console under "Python Server".
+
+### Modifying the LSP
+
+- Clone the [ruff-lsp](https://github.com/charliermarsh/ruff-lsp) to e.g. `../ruff-lsp`
+- Go to `../ruff-lsp` and run `pip install -t ../ruff-vscode/bundled/libs/ -e .`
+- If you want to use a custom ruff build, copy or link it to `bundles/libs/bin/ruff`
--- a/bundled/tool/server.py
+++ b/bundled/tool/server.py
@@ -2,30 +2,68 @@
 
 from __future__ import annotations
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -154,3 +154,12 @@ This extension is based on the [Template for VS Code Python tools extensions](ht
...
 - To run tests, run: `just test`.
+
+To run the extension, navigate to `src/extension.ts` and run (`F5`). You should see the LSP output
+and Python log messages in the debug console under "Python Server".
+
+### Modifying the LSP
+
+- Clone the [ruff-lsp](https://github.com/charliermarsh/ruff-lsp) to e.g. `../ruff-lsp`
+- Go to `../ruff-lsp` and run `pip install -t ../ruff-vscode/bundled/libs/ -e .`
+- If you want to use a custom ruff build, copy or link it to `bundles/libs/bin/ruff`
```
MichaReiser, line 165: You can also change the path in the extension settings

bundled/tool/server.py
```diff
@@ -2,30 +2,68 @@
...
+                    "class": "logging.StreamHandler",
+                    "formatter": "simple",
+                },
+            },
+            "root": {"level": "INFO", "handlers": ["stderr"]},
+            "loggers": {
+                # Don't repeat every message
+                "pygls.protocol": {
+                    "level": "WARN",
+                    "handlers": ["stderr"],
```
MichaReiser, line 49: Is my assumption correct that this writes to stderr and not the output pane?

konstin: It also writes to the output pane just in red instead of blue

MichaReiser: Oh nice. Is this because the vs code extension redirects stderr to the output panel?

konstin: yes, i think they generally just gather all output

charliermarsh:
Will use this opportunity to annoyingly comment on a couple procedural nits:

1. Would prefer to standardize on imperative titles, like "Improve development experience" over "Better development experience".
2. In the summary, would be great to include more on the what and why (e.g., "This PR adds instructions to streamline the local development of Ruff LSP alongside the VS Code extension", and so on).

konstin:
Will

README.md
konstin: good point, changed

Issue #127, opened 2023-02-07
Error: command 'ruff.applyAutofix' already exists


tekumara:
When starting vscode sometimes:

```
Name: Ruff
Module: ruff
Python extension loading
Waiting for interpreter from python extension.
Python extension loaded
Server Command [RUN]: /Users/tekumara/.pyenv/versions/3.9.13/bin/python /Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/bundled/tool/server.py
Server Command [DEBUG]: /Users/tekumara/.pyenv/versions/3.9.13/bin/python /Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/bundled/tool/_debug_server.py
Server Command [RUN]: /Users/tekumara/code/dt-xade-prefect-utils/.venv/bin/python /Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/bundled/tool/server.py
Server Command [DEBUG]: /Users/tekumara/code/dt-xade-prefect-utils/.venv/bin/python /Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/bundled/tool/_debug_server.py
[Error - 15:39:07] Server initialization failed.
Error: command 'ruff.applyAutofix' already exists
	at s.registerCommand (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:96:106489)
	at Object.registerCommand (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:117:29961)
	at Y.register (/Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/dist/extension.js:1:162967)
	at Y.initialize (/Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/dist/extension.js:1:162539)
	at w.initializeFeatures (/Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/dist/extension.js:1:183031)
	at /Users/tekumara/.vscode/extensions/charliermarsh.ruff-2023.6.0-darwin-arm64/dist/extension.js:1:174528
[Error - 15:39:07] Starting client failed
Error: command 'ruff.applyAutofix' already exists
[...]

PR #194, opened 2023-05-16
Only one server restart at a time

konstin:
This hopefully solves #127

I still have to find a better way to test this

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,6 +15,8 @@ import { getProjectRoot } from './common/utilities';
 import { createOutputChannel, onDidChangeConfiguration, registerCommand } from './common/vscodeapi';
 
 let client: LanguageClient | undefined;
+let clientPromise: Promise<LanguageClient | undefined> | undefined = undefined;
+let isNewRestartQueued: boolean = false;
 
 export async function activate(context: vscode.ExtensionContext): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -33,7 +35,28 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
     traceVerbose(`Configuration: ${JSON.stringify(serverInfo)}`);
 
     const runServer = async () => {
-        client = await restartServer(serverId, serverName, outputChannel, client);
+        if (clientPromise) {
```
MichaReiser, line 38: clientPromise != null

Let's not depend on implicit conversion (it's safe here because it is a promise). We should add eslint or Rome

konstin:
I confirmed that the code works but doesn't fix the problem. Not sure what to do with this

src/extension.ts
konstin: Please add the appropriate tooling, i'm not up to date what the appropriate tooling in the js ecosystem is.

konstin: i tried adding eslint and it didn't flag this, i think i got the wrong set of rules

MichaReiser (approved):
I would merge it because it fixes one potential cause for the race and may provide us with additional information to fix it properly.

PR #201, opened 2023-06-10
Exclude build-only artefacts from vsix

### LINKED ISSUE #185: Exclude build-only artefacts from vsix.
grafik

Tests, `build`, `justfile` etc should not be included in the built vsix file.

harupy:
Resolve #185


charliermarsh:
Thanks!

PR #202, opened 2023-06-11
Fix eqeqeq warning

harupy:
Saw this in #201 

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,7 +11,7 @@
         "@typescript-eslint/naming-convention": "warn",
         "@typescript-eslint/semi": "warn",
         "curly": "warn",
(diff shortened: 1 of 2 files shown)


konstin:
This lint warning should be handle, but in this case `!==` doesn't do the right thing either as we initialize `clientPromise` as undefined:
```javascript
let clientPromise = undefined;
console.log(clientPromise != null);
console.log(clientPromise !== null);
```
```
false
true
```
One solution would be intializing `clientPromise` to `null`, but i don't understand js (or ts) best practices enough to make a decision here

charliermarsh:
Can we disable this check for null comparisons? I prefer `!= null` for this reason -- it handles both null and undefined.

PR #233, opened 2023-06-19
Remove unused showNotifications setting

yaegassy:
## Summary

It appears that the setting for `showNotifications` has been removed in the latest version of `ruff-lsp`. Therefore, we have also removed it on the `ruff-vscode` side.

- <https://github.com/astral-sh/ruff-lsp/pull/149>

## Test Plan

The following command has been executed/completed.

- `just fmt`
- `just check`
- `just test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,24 +137,6 @@
           ],
           "default": "off",
           "markdownDescription": "Traces the communication between VSCode and the ruff-lsp."
-        },
-        "ruff.showNotifications": {
-          "default": "off",
-          "description": "Controls when notifications are shown by this extension.",
-          "enum": [
-            "off",
-            "onError",
-            "onWarning",
-            "always"
-          ],
-          "enumDescriptions": [
-            "All notifications are turned off, any errors or warning are still available in the logs.",
-            "Notifications are shown only in the case of an error.",
-            "Notifications are shown for errors and warnings.",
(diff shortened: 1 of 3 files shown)


charliermarsh:
It actually _is_ used! But it's set as an environment variable by the VS Code extension rather than exposed as an LSP setting. I'm not 100% sure why, but it's how the VS Code extension template is setup. The relevant code in the LSP:

```python
def log_error(message: str) -> None:
    LSP_SERVER.show_message_log(message, MessageType.Error)
    if os.getenv("LS_SHOW_NOTIFICATION", "off") in ["onError", "onWarning", "always"]:
        LSP_SERVER.show_message(message, MessageType.Error)


def log_warning(message: str) -> None:
    LSP_SERVER.show_message_log(message, MessageType.Warning)
    if os.getenv("LS_SHOW_NOTIFICATION", "off") in ["onWarning", "always"]:
        LSP_SERVER.show_message(message, MessageType.Warning)


def log_always(message: str) -> None:
    LSP_SERVER.show_message_log(message, MessageType.Info)
    if os.getenv("LS_SHOW_NOTIFICATION", "off") in ["always"]:
        LSP_SERVER.show_message(message, MessageType.Info)
```

yaegassy:
Ah, I see.

PR #251, opened 2023-07-06
Add some scrappy scripts to facilitate version bumps

charliermarsh:


scripts/bump_extension_version.sh
```diff
@@ -0,0 +1,31 @@
...
+# Re-lock dependencies.
+rm requirements.txt
+rm requirements-dev.txt
+pip-compile --generate-hashes --resolver=backtracking -o ./requirements.txt ./pyproject.toml
+pip-compile --generate-hashes --resolver=backtracking --upgrade --extra dev -o ./requirements-dev.txt ./pyproject.toml
+npm install --package-lock-only
+
+# Commit the change.
+git add .
+git commit -m "Bump extension version to $TO"
```
zanieb, line 28: nit: I'd use `git commit -am` instead of `git add .` unless you _want_ to add untracked files to the commit. Otherwise it seems too easy to include files by accident.

charliermarsh: Thanks!

PR #273, opened 2023-09-02
Add a VS Code setting to enable the experimental formatter

charliermarsh:
## Summary

This PR adds an `enableExperimentalFormatter` flag to the extension which turns on formatting capabilities by setting the environment variable in the LSP server. Our extension is setup such that if you change this setting, we restart the server, so it is respected across modifications. This is (intentionally) more discoverable, as I had trouble with the environment variable alone.

## Test Plan

- Run the extension locally.
- Set `"editor.defaultFormatter": "charliermarsh.ruff"` in `settings.json`.
- Try to format; verify that Ruff is not a valid option.
- Enable the setting (`"ruff.enableExperimentalFormatter": true`).
- Try to format; verify that it works as expected.
- Disable the setting; try to format; verify that it fails as expected:

Screen Shot 2023-09-02 at 5 23 25 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,6 +193,12 @@
           ],
           "scope": "machine",
           "type": "string"
+        },
+        "ruff.enableExperimentalFormatter": {
+          "default": false,
+          "markdownDescription": "Controls whether Ruff registers as capable of code formatting. The Ruff formatter is in an alpha state during which formatting may change at any time.",
+          "scope": "machine",
+          "type": "boolean"
         }
       }
     },
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -44,6 +44,11 @@ async function createServer(
   // Set notification type
   newEnv.LS_SHOW_NOTIFICATION = settings.showNotifications;
 
+  // Set experimental formatter capabilities
+  if (settings.enableExperimentalFormatter) {
+    newEnv.RUFF_EXPERIMENTAL_FORMATTER = "1";
+  }
+
   const args =
     newEnv.USE_DEBUGPY === "False" || !isDebugScript
       ? settings.interpreter.slice(1).concat([SERVER_SCRIPT_PATH])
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -193,6 +193,12 @@
           ],
           "scope": "machine",
           "type": "string"
+        },
+        "ruff.enableExperimentalFormatter": {
+          "default": false,
+          "markdownDescription": "Controls whether Ruff registers as capable of code formatting.",
```
MichaReiser, line 199: I think we should be more explicit about what experimental means in this context and whether toggling the setting requires restarting VS code.

src/common/server.ts
```diff
@@ -44,6 +44,11 @@ async function createServer(
   // Set notification type
   newEnv.LS_SHOW_NOTIFICATION = settings.showNotifications;
 
+  // Set experimental formatter capabilities
+  if (settings.enableExperimentalFormatter) {
+    newEnv.RUFF_EXPERIMENTAL_FORMATTER = "1";
+  }
```
MichaReiser, line 50: Does this require restarting VS Code after toggling the setting? If so, we should document this in the setting.

MichaReiser (approved):
Looks good. The downside of this approach over dynamic capability registration is that it only works for VS Code. Meaning, each editor needs to implement the same handling. 

@zanieb we may need to figure out what enabling preview features means for IDE extensions. Should the extension read the `ruff.toml` and or respect the `--preview` CLI argument? This is another use case where it would be easier if ruff-lsp is written in Rust and hosts ruff... Maybe a project for leap weak.

src/common/server.ts
charliermarsh: No, it doesn't require restarting VS Code in my testing. (See the test plan above for a description of what it looks like in practice.)

PR #282, opened 2023-09-19
Update README.md with black and URLs

pamelafox:
## Summary

My original goal was just to update the deprecated setting python.formatting.provider (per https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions) but then I also updated a bunch of charliemarsh urls to astral-sh.

## Test Plan

Click links, see result in window.

For setting change, I checked with VS code team and will also test in profile/devcontainer.


charliermarsh:
Thanks Pamela!

PR #283, opened 2023-09-20
Update README.md with example for using `ruff.args` in `settings.json`

simkimsia:
Based on a suggestion in https://github.com/astral-sh/ruff/discussions/7474#discussioncomment-7042823, I'm making this change to make it easier for newbies like me to customize ruff using settings.json.

Specifically using `ruff.args` in `settings.json`

## Summary

As a newbie i didn't know i can use `ruff.args`. see https://github.com/astral-sh/ruff/discussions/7474#discussioncomment-7042823. So I added an example in the vscode extension readme.

## Test Plan

This is README, so no tests needed


fa-ribeiro:
Note that VS Code Insiders (Version: 1.84.0-insider) is marking `ruff.args` as deprecated and advising to use `ruff.lint.args` instead.

simkimsia:
I have noticed this just yesterday on regular vs code

Will change this when I reach laptop in 1 hr

charliermarsh (approved):
Thanks, sorry for the delay.

PR #293, opened 2023-10-02
Move `args` and `run` settings to `lint.args` and `lint.run`

charliermarsh:
## Summary

This is the peer PR to https://github.com/astral-sh/ruff-lsp/pull/255, which adds `lint.args` and `lint.run` to the VS Code extension, and deprecates `args` and `run`:

Screen Shot 2023-10-02 at 3 07 32 PM

The behavior is as follows:

- If you have `lint.run` defined in your settings, we will respect it and ignore `run`.
- Otherwise, if you have `run` defined in your settings, we will respect it.
- Otherwise, we fall back to the default value.

Confusingly, if you set `run`, VS Code _still_ doesn't show it in the visual settings UI, which is unfortunate, as one edge case here is:

- You have `run` defined.
- You set `lint.run` in the visual UI.
- (`run` is still defined in your `settings.json`, but not visible in the UI.)
- You unset `lint.run` in the visual UI.
- (VS Code then removes `lint.run` from your `settings.json`, since it now matches the default value and is unnecessary.)
- Now, you're using the old `run` setting even though it's not visible in the UI.

I'm not sure if there's a great way around this. We could consider showing a warning, but that might be somewhat intrusive?

## Test Plan

Extensive local testing 😅

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -59,19 +59,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
   traceLog(`Module: ${serverInfo.module}`);
   traceVerbose(`Full Server Info: ${JSON.stringify(serverInfo)}`);
 
+  context.subscriptions.push(
+    onDidChangeConfiguration((event) => {
+      if (event.affectsConfiguration("ruff.enable")) {
+        vscode.window.showWarningMessage(
+          "To enable or disable Ruff after changing the `enable` setting, you must restart VS Code.",
+        );
+      }
+    }),
+  );
+
   const { enable } = getConfiguration(serverId) as unknown as ISettings;
   if (!enable) {
     traceLog(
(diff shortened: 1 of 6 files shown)


charliermarsh:
I'm not sure why I merged https://github.com/astral-sh/ruff-vscode/pull/292 which was the half-finished version of this PR. I must've been barely awake.

src/extension.ts
```diff
@@ -59,20 +59,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
 
+  context.subscriptions.push(
+    onDidChangeConfiguration((event) => {
+      if (event.affectsConfiguration("ruff.enable")) {
+        vscode.window.showWarningMessage(
+          "To enable or disable Ruff after changing the `enable` setting, you must restart VS Code.",
+        );
+      }
+    }),
+  );
```
charliermarsh, line 70: Unrelated, but this now prints as a popup rather than in the trace log.

PR #297, opened 2023-10-11
Add missing "s" in deprecation message

lyzlisa:
## Summary

When updating from `ruff.args` to new syntax `ruff.lint.args`, the helper message indicated to use `ruff.lint.arg` which was missing an `s` at the end, which in turn showed "Unknown Configuration Setting".

All other references to new syntax seem to be right, except for https://github.com/astral-sh/ruff-vscode/blame/be792045f904bb711e6cc29895a6d1ed8dfe0440/package.json#L68

Screenshot 2023-10-11 at 10 42 20
Screenshot 2023-10-11 at 10 42 48


zanieb:
Thanks!

PR #322, opened 2023-11-01
chore: move `eslint` and `prettier` configuration within `package.json`

SauravMaheshkar:
## Summary

Changes proposed by this PR can be summarised as follows :-

* Move the content from `.prettierrc.json` under the `prettier` key in `package.json`. [Source(prettier docs)](https://prettier.io/docs/en/configuration.html)
* Move the content from `.eslintrc.json` and `.eslintignore` under the `eslintConfig` and `eslintIgnore` keys respectively in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files)

This will lead to a minimal repository structure and improve overall DX


zanieb (approved):
Sounds good to me

charliermarsh:
Thanks!

PR #326, opened 2023-11-03
Update with Notebook support

dhruvmanila:
We should probably wait until `ruff-lsp` is released before we merge this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The extension ships with `ruff==0.1.2`.
 ## Usage
 
 Once installed in Visual Studio Code, `ruff` will automatically execute when you open or edit a
-Python file.
+Python or Jupyter Notebook file.
 
 If you want to disable Ruff, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension)
 per workspace in Visual Studio Code.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,19 @@ using "Fix all" can be enabled by setting `unsafe-fixes = true` in your Ruff con
...
 safety works.
 
+## Jupyter Notebook Support
+
+The extension has support for Jupyter Notebooks via the [Notebook Document Synchronization] capabilities of the Language
+Server Protocol which were added in 3.17. This has been implemented in `ruff-lsp` as of version `v0.0.43` which provides
+full support for all of the existing capabilities available to Python files in Jupyter Notebooks, including diagnostics,
+code actions, and formatting.
+
+This requires Ruff version `v0.1.3` or greater. Refer to the [Jupyter Notebook Discovery] section to learn how to configure
```
charliermarsh, line 63: Nit: we typically say "or later" rather than "or greater".

README.md
```diff
@@ -53,6 +53,19 @@ using "Fix all" can be enabled by setting `unsafe-fixes = true` in your Ruff con
...
+Ruff to work with Jupyter Notebooks.
```
charliermarsh, line 64: Do they need to change this to get the VS Code extension to work? Doesn't it work without this, since we pass the file directly?

README.md
```diff
@@ -125,10 +169,30 @@ following `settings.json`:
...
+```
+
 _Note: if you're using Ruff to organize imports in VS Code and _also_ expect to run Ruff from the
 command line, you'll want to enable Ruff's isort rules by by adding `"I"` to your
 [`extend-select`](https://docs.astral.sh/ruff/settings/#extend-select)._
 
+_Note: All of the above mentioned Notebook configurations will run the action for each cell individually.
+This is the way VS Code handles Notebook actions and is unrelated to `ruff-lsp`. If you'd prefer to run
+them on the entire notebook at once, prefer to use the `Ruff` prefixed commands such as
+`Ruff: Organize Imports` and `Ruff: Fix all auto-fixable problems`._
```
charliermarsh, line 194: This isn't true for "Format document" though, right? (At least, it wasn't in my testing.)

charliermarsh (approved):

README.md
dhruvmanila: Oh, right. I'll remove that part. Thanks for noticing!

README.md
dhruvmanila: It is true. Here, it sent 2 requests for a notebook containing 2 cells:
```
2023-11-08 08:20:15.130 [info] [Trace - 8:20:15 AM] Sending request 'textDocument/formatting - (15)'.
2023-11-08 08:20:15.130 [info] [Trace - 8:20:15 AM] Sending request 'textDocument/formatting - (16)'.
```

Even though VS Code runs them for each cell, they run it in parallel. This will seem like it performed the action on an entire Notebook.

PR #375, opened 2023-12-29
Update ruff README badge versions to 0.1.9

kieran-ryan:
## Summary

Current ruff versions in README badges use `0.1.0`...

[image](https://pypi.python.org/pypi/ruff)
[image](https://pypi.python.org/pypi/ruff)
[image](https://pypi.python.org/pypi/ruff)

...instead of `0.1.9`:

[image](https://pypi.python.org/pypi/ruff)
[image](https://pypi.python.org/pypi/ruff)
[image](https://pypi.python.org/pypi/ruff)

Looks like missed in bump from `0.1.0` to `0.1.2` (#309).

## Test Plan

Updated the badges to `0.1.9` and then executed the script to bump the versions.

```console
./scripts/bump_ruff_version.sh 0.1.9 0.1.10
```

All versions - including those in the README badges - updated successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # Ruff extension for Visual Studio Code
 
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
-[![image](https://img.shields.io/pypi/v/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
-[![image](https://img.shields.io/pypi/l/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
-[![image](https://img.shields.io/pypi/pyversions/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
+[![image](https://img.shields.io/pypi/v/ruff/0.1.9.svg)](https://pypi.python.org/pypi/ruff)
+[![image](https://img.shields.io/pypi/l/ruff/0.1.9.svg)](https://pypi.python.org/pypi/ruff)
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks!

README.md
```diff
@@ -1,9 +1,9 @@
 # Ruff extension for Visual Studio Code
 
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
-[![image](https://img.shields.io/pypi/v/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
-[![image](https://img.shields.io/pypi/l/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
-[![image](https://img.shields.io/pypi/pyversions/ruff/0.1.0.svg)](https://pypi.python.org/pypi/ruff)
+[![image](https://img.shields.io/pypi/v/ruff/0.1.9.svg)](https://pypi.python.org/pypi/ruff)
+[![image](https://img.shields.io/pypi/l/ruff/0.1.9.svg)](https://pypi.python.org/pypi/ruff)
+[![image](https://img.shields.io/pypi/pyversions/ruff/0.1.9.svg)](https://pypi.python.org/pypi/ruff)
```
dhruvmanila, line 6: We don't really need to include the version as it'll automatically pick up the latest one. So, the link would look like https://img.shields.io/pypi/v/ruff (same for others).

kieran-ryan: @dhruvmanila, had considered; thinking was that the extension ships with a versioned distribution of ruff - and if there’s a ruff release with breaking changes for the extension, users would see that version in the extension marketplace if the latest was used and could assume compatibility and raise a bug. What do you think?

Happy with either, just a consideration. Though certainly appears an improvement is needed, as at present they appear to go out of sync if the script to bump the version is not run to update the version e.g. if manually updated.

PR #380, opened 2024-01-05
Amend `notebook.source` actions on save recommendation

blakeNaccarato:
Looks like this line got prefixed with `notebook` despite the setting targeting non-notebook files. Just a one-line change, so I hope it's alright I didn't raise an issue on it. Fixes what looks like a one-line error introduced in https://github.com/astral-sh/ruff-vscode/pull/379?

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Similarly, you can configure Ruff to organize imports on-save by enabling the
 {
   "[python]": {
     "editor.codeActionsOnSave": {
-      "notebook.source.organizeImports": true
+      "source.organizeImports": true
     }
   }
 }


blakeNaccarato:
Also, I was tempted to swap `true` for `"always"` in `editor.codeActionsOnSave` entries and `"explicit"` in `notebook.codeActionsOnSave` entries, (see recommendations [here](https://code.visualstudio.com/updates/v1_83#_code-actions-on-save-and-auto-save) and [here](https://code.visualstudio.com/updates/v1_83#_finalized-notebook-code-action-api)), but wanted to limit the changeset since `true` is still technically allowed.

charliermarsh:
Oops, thank you!

zanieb (approved):
LGTM! Thank you for contributing!

PR #385, opened 2024-01-15
support more architectures

diceroll123:
## Summary

Support a few more architectures

## Test Plan

I have not tested it. Ideally the PR will do a dry run.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -68,6 +68,17 @@ jobs:
           - os: macos-latest-xlarge
             target: aarch64-apple-darwin
             code-target: darwin-arm64
+          - os: ubuntu-20.04
+            target: x86_64-unknown-linux-musl
+            code-target: linux-x64
+          - os: ubuntu-20.04
+            target: aarch64-unknown-linux-musl
+            code-target: linux-arm64
+            arch: aarch64
+          - os: ubuntu-20.04
+            target: armv7-unknown-linux-musleabihf
+            code-target: linux-armhf
+            arch: armv7
 
     name: Build (${{ matrix.target }})
     needs: ["build-id"]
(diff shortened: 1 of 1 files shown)


diceroll123:
Guess not!

konstin:
We have to manually approve CI runs each times for first-time contributors unfortunately. You can activate github actions in your fork and make a PR in your fork to test.

diceroll123:
Updated.

[In my PR on my fork](https://github.com/diceroll123/ruff-vscode/pull/1), I:
- made it so release workflow runs on PR, for testing purposes
- disabled macos builds because they cost money apparently
- removed the build ID and publish steps because I didn't need them

But the only changes I've brought over to here are the build architectures. Attached is the list of successful actions!

image

`aarch64-unknown-linux-musl` I believe is what will allow us to use ruff within a devcontainer running on alpine linux, can't wait! 😄

charliermarsh:
I can do some testing and ultimately build for nightly, thanks :)

.github/workflows/release.yaml
```diff
@@ -68,6 +68,17 @@ jobs:
...
+            target: x86_64-unknown-linux-musl
+            code-target: linux-x64
+          - os: ubuntu-20.04
+            target: aarch64-unknown-linux-musl
+            code-target: linux-arm64
+            arch: aarch64
+          - os: ubuntu-20.04
+            target: armv7-unknown-linux-musleabihf
+            code-target: linux-armhf
+            arch: armv7
```
charliermarsh, line 81: I think this isn't quite right because these `code-target` values are the same as the ones above. We need to support `alpine-x64` and `alpine-arm64`. See: https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions.

charliermarsh: I need to play around with things (https://github.com/astral-sh/ruff-vscode/pull/387) to understand how this works. I think we need the extension itself to be built under Alpine.

charliermarsh: @diceroll123 -- Okay, I think I got this working [here](https://github.com/astral-sh/ruff-vscode/pull/387/files) -- notice how the build is done within an Alpine container, does that make sense?

diceroll123: Absolutely! Awesome stuff. 😄 Learning a lot from ya.

charliermarsh: Thanks for doing this, and sorry that it didn't get merged!

PR #389, opened 2024-01-15
Propagate build ID task error

charliermarsh:
## Summary

See: https://github.com/astral-sh/ruff-vscode/actions/runs/7533695396/job/20506697608. If the commands fail, the step should fail.


charliermarsh:
I can't get this to work, sigh.

Issue #329, opened 2023-11-06
Allow using the formatter but not the linter


Apakottur:
I'm using the Ruff formatter in VSCode, but would prefer to run the linter manually outside of it.
The current configuration options do not provide a way to enable the formatter but disable the linter, so I use the following hack to make it happen:

```json
...
"ruff.lint.args": [
        "--crash"
    ],
...
```

Am I missing something, and it's possible to disable the linter but not the formatter?

dhruvmanila:
Thanks for the clear issue!

I don't think it's currently possible to disable just the linter as it's provided through the LSP interface where the diagnostics are always provided. Maybe we could provide a configuration option in `ruff-lsp` to disable diagnostics.

If possible, can you provide some details as to why you would want to run the linter outside the editor? Your use-case might help us understand this issue more clearly.

charliermarsh:
I don't think this is supported but I did consider adding it in the past. (You could probably change `--crash` to `--ignore=ALL`.)

Apakottur:
Thanks for the replies.

We use a very strict Ruff configuration in the project I'm working on (almost all rules enabled), so Ruff can be very noisy when writing code. For example when adding an import (which is immediately unsorted) or introducting a new variable (which is immediately unused).

Until recently we've been using the Ruff linter and the Black formatter, and I used Black in VSCode. Recently we switched to the Ruff formatter so I wanted to use the Ruff formatter in VSCode.

I realize that I might create a proprietary configuration file for the VSCode Ruff linter (in contrast to the strict one we use in pre-commit and CI), but right now I don't feel like running the Ruff linter more often than I do will increase productivity.

T-256:
I like it. Extra `Ruff: Toggle linting` command or something like [Formatting Toggle](https://marketplace.visualstudio.com/items?itemName=tombonnike.vscode-status-bar-format-toggle) could help to get rid from external terminal ruff execution.

`Ruff: Toggle linting`: focus on writing codes without linter warnings for unfinished lines
few hours coding...
`Ruff: Toggle linting` again: now fix linter violations (if any) on finished codes.

zpincus:
Not unrelated, it would be nice to be able to toggle linting for notebooks separately from python source...

PR #394, opened 2024-01-24
Add support for using the extension exclusively for formatting

charliermarsh:
## Summary

This PR enables users to specify `"ruff.lint.enable": false` to use the LSP (or extension) solely for formatting.

"Organize Imports" is still supported even when the linter is disabled.

Closes: https://github.com/astral-sh/ruff-vscode/issues/329.

Depends on: https://github.com/astral-sh/ruff-lsp/pull/361.

## Test Plan

- Ran the VS Code extension.
- Verified that diagnostics appeared for unused imports.
- Set `"ruff.lint.enable": false`.
- Verified that the diagnostics disappeared.
- Ran "Organize Imports".
- Verified that the imports were sorted.
- Ran "Fix All".
- Verified that the imports (which were unused) were not removed.


pyproject.toml
```diff
@@ -56,3 +42,82 @@ module = [
   "pylsp_jsonrpc.*",
 ]
 ignore_missing_imports = true
+
+[tool.ruff]
```
dhruvmanila, line 46: Is this change relevant for this PR? There's also the `ruff.toml` file which probably contains the same settings as mentioned here.

dhruvmanila (approved):

pyproject.toml
charliermarsh: Thanks, these were stupidly checked-in from when I was doing some testing of our JSON Schema...

README.md
```diff
@@ -67,17 +67,18 @@ This requires Ruff version `v0.1.3` or later.
...
+| interpreter                          | `[]`              | Path to a Python interpreter to use to run the linter server.                                                                                                                                                                                                     |
+| lint.args                            | `[]`              | Additional command-line arguments to pass to `ruff check`, e.g., `"args": ["--config=/path/to/pyproject.toml"]`. Supports a subset of Ruff's command-line arguments, ignoring those that are required to operate the LSP, like `--force-exclude` and `--verbose`. |
+| line.enable                          | `true`            | Whether to enable linting. Set to `false` to use Ruff exclusively as a formatter.                                                                                                                                                                                 |
```
MichaReiser, line 78: Shouldn't this be `lint.enable`?

charliermarsh: No! no no no!

PR #406, opened 2024-02-09
Update ruff version in README.md

lociii:
## Summary

Update ruff version in README according to the latest one supported in 2024.10.0

## Test Plan

Not necessary


charliermarsh:
Thanks!

PR #429, opened 2024-03-19
fix typo in README for the ruff.showNotifications setting

grisaitis:
## Summary

hi! fixed a simple typo in the root README. `showNotification` should be `showNotifications`. i double checked this in the source elsewhere. discovered the issue when adjusting the vscode setting myself, and reading the README.

## Test Plan

n/a


charliermarsh:
Thanks!

PR #436, opened 2024-03-28
Remove warning about restarting VS Code that appeared when enabling/disabling the experimental server

snowsignal:
## Summary

We don't actually need to tell the user to restart VS Code when changing this setting. I put this in as a precautionary measure but from the testing I've done there doesn't seem to be a need to restart it.

## Test Plan

https://github.com/astral-sh/ruff-vscode/assets/19577865/f1536261-abc7-4016-89c9-c61e9c4047a6

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -66,10 +66,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         vscode.window.showWarningMessage(
           "To enable or disable Ruff after changing the `enable` setting, you must restart VS Code.",
         );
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Would you mind a short test video as test plan that shows switching between the two? It can be a simple no-sound walkthrough (nothing fancy). Having a test plan allows me to verify the change without having to checkout the PR myself (and I can see the flows you've tested and I can think about if there are any other flows that need testing)

snowsignal:
@MichaReiser I've added a test plan video, as requested.

PR #443, opened 2024-04-10
Experimental `ruff server` now uses local `ruff` binaries when available 

snowsignal:
## Summary

When `experimentalServer` is enabled, the Ruff binary path will be determined in the following order:

1. **`ruff.path`** - if this is set, we take the first path in the list and use that as the executable path.
2. **`sysconfig.get_path("scripts") / RUFF_EXE`** -  `RUFF_EXE` is `"ruff.exe"` on windows and `"ruff"` otherwise.
3. **`shutil.which("ruff")`**
4. The bundled binary path, **`./bundled/libs/bin/ruff`**, is used if none of the prior options are available.

Before running the binary, we also check to make sure that it's running a compatible version. At the moment, any version after or including `0.3.3` is supported (except for `0.3.4`, since that release had a major bug in the server that broke editor integration). Since `0.3.3` is a rather bare-bones implementation, this minimum supported version will increase as the extension stabilizes.

## Test Plan

1. Run `ruff --version` in your shell. If it's not `ruff 0.3.5` or later, install/upgrade `ruff` and run `ruff --version` again to confirm that the ruff executable in your `PATH` is now `>=0.3.5`.
2. Begin debugging the extension. You should see the following log somewhere in the output: `Configuration file watcher successfully registered`. This should be a confirmation that you're running `ruff >=0.3.5` instead of the embedded `ruff 0.3.3`.
3. Add an entry to `ruff.path` in the extension settings, and pass in a ruff binary not in your `PATH`. A good way to check that you are in fact using this unique binary is by passing in the path to a local debug build after adding an additional log statement in a function like `Server::new`. If this log statement appears in the output, you know that you're using the custom executable provided in `ruff.path`. 
4. Now, uninstall the ruff executable that was in your `PATH`. Confirm that no executable exists in your path by running `ruff --version` in your shell - you should get your shell's variant of the `command not found: ruff` error.
5. At this point, the extension should still be using the custom executable you added to `ruff.path`. Now, remove that path so that `ruff.path` is an empty list. The extension should now be using the bundled `v0.3.3` binary (a quick way to check this is making sure no `Configuration file watcher successfully registered` message appears, though you could also confirm this by seeing if the diagnostics show up highlighted in red instead of yellow).


bundled/tool/ruff_server.py
```diff
@@ -0,0 +1,89 @@
...
+                "Please upgrade to the latest version and try again."
+            )
+            raise RuntimeError(message)
+
+
+def find_ruff_bin(fallback: Path) -> Path:
+    """Return the ruff binary path."""
+    path = Path(sysconfig.get_path("scripts")) / RUFF_EXE
+    if path.is_file():
+        return path
```
charliermarsh, line 72: I think we might as well change this to use the `find_ruff_bin` logic that we use in Ruff itself:

```python
def find_ruff_bin() -> str:
    """Return the ruff binary path."""

    ruff_exe = "ruff" + sysconfig.get_config_var("EXE")

    path = os.path.join(sysconfig.get_path("scripts"), ruff_exe)
    if os.path.isfile(path):
        return path

    if sys.version_info >= (3, 10):
        user_scheme = sysconfig.get_preferred_scheme("user")
    elif os.name == "nt":
        user_scheme = "nt_user"
    elif sys.platform == "darwin" and sys._framework:
        user_scheme = "osx_framework_user"
    else:
        user_scheme = "posix_user"

    path = os.path.join(sysconfig.get_path("scripts", scheme=user_scheme), ruff_exe)
    if os.path.isfile(path):
        return path

    raise FileNotFoundError(path)
```

(Like, don't call Ruff, just add this function here.)

bundled/tool/ruff_server.py
```diff
@@ -0,0 +1,89 @@
...
+
+if __name__ == "__main__":
+    ruff = os.fsdecode(
+        find_ruff_bin(
+            Path(BUNDLE_DIR / "libs" / "bin" / RUFF_EXE),
+        ),
+    )
+    check_compatibility(ruff, RUFF_VERSION_REQUIREMENT, FORBIDDEN_RUFF_VERSIONS)
+    completed_process = subprocess.run([ruff, *sys.argv[1:]], check=False)
+    sys.exit(completed_process.returncode)
```
charliermarsh, line 89: In Ruff, we do:

```python
if __name__ == "__main__":
    ruff = os.fsdecode(find_ruff_bin())
    if sys.platform == "win32":
        import subprocess

        completed_process = subprocess.run([ruff, *sys.argv[1:]])
        sys.exit(completed_process.returncode)
    else:
        os.execvp(ruff, [ruff, *sys.argv[1:]])
```

Can we use that same logic here?

src/common/constants.ts
```diff
@@ -10,6 +10,10 @@ export const DEBUG_SERVER_SCRIPT_PATH = path.join(
   "tool",
   `_debug_server.py`,
 );
-export const RUFF_BIN_PATH = path.join(BUNDLED_PYTHON_SCRIPTS_DIR, "libs", "bin", "ruff");
+export const NEW_SERVER_SCRIPT_PATH = path.join(
```
charliermarsh, line 13: I'd vote `EXPERIMENTAL_SERVER_SCRIPT_PATH`, if that's what you mean by `NEW_`, so it's consistent with the setting? Easier to find later.

src/common/server.ts
```diff
@@ -39,17 +39,31 @@ async function createExperimentalServer(
...
-  };
+    serverOptions = {
+      command,
+      args,
+      options: { cwd, env: process.env },
+    };
 
-  traceInfo(`Server run command: ${[command, ...args].join(" ")}`);
+    traceInfo(`Server run command: ${[command, ...args].join(" ")}`);
+  }
```
charliermarsh, line 66: Can you add a comment to each branch here, like:

```js
// If the user provided a path to an explicit binary, ...
```

bundled/tool/ruff_server.py
```diff
@@ -0,0 +1,89 @@
...
+    # Ensure that we can import bundled libraries like `packaging`
+    update_sys_path(os.fspath(BUNDLE_DIR / "libs"))
+
+
+from packaging.specifiers import SpecifierSet
+from packaging.version import Version
+
+# This is the first release that included `ruff server`.
+# The minimum version may change in the future.
+RUFF_VERSION_REQUIREMENT = SpecifierSet(">=0.3.3")
```
charliermarsh, line 36: For simplicity, should we just make this `0.3.5` or later?

charliermarsh:
It looks like the only things this is missing vis-a-vis the existing LSP (not that they're required for the Alpha, or even at all -- just confirming my understanding) are:

1. `importStrategy` -- allow the user to prioritize the bundled version.
2. `interpreter` -- allow the user to specify an alternative Python interpreter.

snowsignal:
> `interpreter` -- allow the user to specify an alternative Python interpreter.

The `interpreter` setting should be supported (we read from it [here](https://github.com/astral-sh/ruff-vscode/pull/443/files#diff-257a47cc991fba869a2e739135efe53969820acc8d326045bfc1de8ee4eba793R55)), but we only read from the first path in the list, if it exists.

You're right that we aren't using `importStrategy` yet. Another limitation with this implementation is that it doesn't use subsequent entries in `ruff.path` past the first one as fallbacks.

bundled/tool/ruff_server.py
snowsignal: This was actually how the code looked when I first ported it over - I can revert it, but I would like to understand why we can't simplify this to a single `subprocess.run` call.

(Ruff threw [S606](https://docs.astral.sh/ruff/rules/start-process-with-no-shell/) for our usage of `os.execvp`, recommending that we use `subprocess` instead)

charliermarsh: I'd suggest looking back at the Git history on that file. I don't know the answer off-hand but I suspect there's more context in the history.

bundled/tool/ruff_server.py
snowsignal: I thought it would be awkward to implement this feature if we only supported the latest version 😅 

But we probably should just make that the minimum, yeah.

snowsignal: We'll also have to bump the bundled Ruff version to make this work (it's at `0.3.3` right now)

bundled/tool/ruff_server.py
snowsignal: It looks like this was added to improve responsiveness on startup: https://github.com/astral-sh/ruff/commit/158367bf913f5a5df2c36f0d160ac35f96b5364c.

Which honestly doesn't seem that necessary in this case, since we aren't repeatedly starting up the server. I'd prefer to keep this as a simple `subprocess.run` invocation.

MichaReiser:
I think we should always use the bundled version if the workspace is not a trusted workspace to avoid executing "untrusted" code:

* `ruffBin`: Could be set in the workspace configuration and point to a file local in the workspace
* `sysconfig`: Could point to a virtual env in the workspace
* `shutil.which`: Could resolve to a file local to the workspace. 
* bundled: That should be safe

src/common/server.ts
```diff
@@ -39,17 +39,34 @@ async function createExperimentalServer(
...
-  const serverOptions: ServerOptions = {
-    command,
-    args,
-    options: { cwd, env: process.env },
-  };
-
-  traceInfo(`Server run command: ${[command, ...args].join(" ")}`);
+  let serverOptions: ServerOptions;
+  // If the user provided a binary path, we'll try to call that path directly.
+  if (settings.path.length > 0 && settings.path[0]) {
```
MichaReiser, line 44: Nit: I think this would be slightly more idiomatic in JS because it's "safe" to access out of bound elements
```suggestion
  if (settings.path[0] != null) {
```

src/common/server.ts
```diff
@@ -39,17 +39,34 @@ async function createExperimentalServer(
...
+      command,
+      args,
+      options: { cwd, env: process.env },
+    };
+
+    traceInfo(`Server run command: ${[command, ...args].join(" ")}`);
+  } else {
+    // Otherwise, we'll call a Python script that tries to locate
+    // a binary, falling back to the bundled version if no local executable is found.
+    const command = settings.interpreter[0];
```
MichaReiser, line 58: I'm happy to tackle this as a separate PR but would it be worth falling back to the bundled version (or using `which` if python isn't installed? 

What I'm thinking of is:

1. Use user provided path
2. If python is installed: Detect from `sysconfig` (requires Python, implemented as Python script)
3. Use which 
4. Fallback to bundled version

I'm leaning towards moving as much as possible to JS, that includes the version check, and only use the python script where we need it to get python's module resolution logic. (That means, the Python script would return the path to a possible ruff binary without performing any of the actual checks).

MichaReiser (approved):
I think it would be good to handle the "untrusted" workspace case before the release but we can do it as a follow up pr. 

A future improvement could be to handle the case where no python installation was detected.

bundled/tool/ruff_server.py
```diff
@@ -0,0 +1,102 @@
...
+        sys.path.insert(0, path_to_add)
+
+        # Allow development versions of libraries to be imported.
+        site.addsitedir(path_to_add)
+
+
+if __name__ == "__main__":
+    # Ensure that we can import bundled libraries like `packaging`
+    update_sys_path(os.fspath(BUNDLE_DIR / "libs"))
+
```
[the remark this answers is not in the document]
charliermarsh: (I believe this _is_ necessary because we need to do this `sys.path` modification _before_ the `packaging` import below.)

charliermarsh:
> I think it would be good to handle the "untrusted" workspace case before the release but we can do it as a follow up pr.

I would comfortably say this isn't required for the release, since it's not something we do today in the current LSP, and the focus on the milestone is to achieve parity with the existing LSP.

(It's also not something that's done in any of the official Microsoft Python template-based extensions. Those don't use `which`, but they do support a direct path, discovery from the interpreter, etc.)

bundled/tool/ruff_server.py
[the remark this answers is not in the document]
snowsignal: This is necessary (for the reason @charliermarsh mentioned) but I'll add a comment explaining this.

src/common/server.ts
snowsignal: Oh, cool! I was not aware of that.

src/common/server.ts
snowsignal: I think these are good ideas, but we should probably explore them in follow-up PRs. I would definitely like to move as much as possible to JS, to the point where we could drop our Python dependency in the future.

bundled/tool/ruff_server.py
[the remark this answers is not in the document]
snowsignal: Actually, would it be OK to just make this a top-level statement? The import is top-level, so even if we import this from elsewhere we'll still need to add `libs` to the system path.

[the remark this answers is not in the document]
charliermarsh: I think it would be best-practice to leave it as-is. If this is imported from elsewhere, it would be uncommon for it to modify `sys.path` IMO.

bundled/tool/ruff_server.py
charliermarsh: That's fine, let me know if you need help doing so.

bundled/tool/ruff_server.py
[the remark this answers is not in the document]
snowsignal: Wouldn't it fail to import if `packaging` wasn't in the system path, though?

charliermarsh (approved):
LGTM though I think you wanted to bump the Ruff version too.

bundled/tool/ruff_server.py
```diff
@@ -0,0 +1,92 @@
+import os
```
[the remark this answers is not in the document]
snowsignal: Yes, our eventual plan is to move this logic to Typescript 😄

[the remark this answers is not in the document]
T-256: Created https://github.com/astral-sh/ruff-vscode/issues/479 to track it.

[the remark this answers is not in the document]
charliermarsh: Would be an awesome change.

PR #444, opened 2024-04-11
Bump version to `2024.19.0-dev`

snowsignal:
## Summary

Bumps the version to the first Alpha pre-release, `2024.19.0-dev`. A new changelog entry has also been added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ruff",
   "displayName": "Ruff",
   "description": "A Visual Studio Code extension with support for the Ruff linter.",
-  "version": "2024.19.0-alpha",
+  "version": "2024.19.0-dev",
   "serverInfo": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ruff",
   "displayName": "Ruff",
   "description": "A Visual Studio Code extension with support for the Ruff linter.",
-  "version": "2024.17.0-dev",
+  "version": "2024.19.0-alpha",
```
charliermarsh, line 5: I think this still needs to be `-dev`.

charliermarsh: Possibly not, trying to read the `update_ext_version.py` code. I'd leave it as `-dev` though, it doesn't end up in any user-facing locations.

charliermarsh (approved):

PR #456, opened 2024-04-21
Introduce common Ruff configuration options to extension settings

snowsignal:
## Summary

This is a follow-up to https://github.com/astral-sh/ruff/pull/10984, which implemented support for common Ruff configuration options in client settings. This PR exposes these new settings in the extension configuration.

These settings are different from other extension settings, because they don't have default values. Instead, if the user does not set them, they will be sent as `null`, and the server will use project configuration instead.

At the moment, `ruff server` does not actually resolve these settings yet - settings resolution is introduced in https://github.com/astral-sh/ruff/pull/11062, which means that you'll need to work from that branch to test this PR (and vice-versa).

## Test Plan

See the test plan in https://github.com/astral-sh/ruff/pull/11062.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,34 +90,40 @@
           "type": "array"
         },
         "ruff.lint.preview": {
-          "markdownDescription": "Enable preview mode for the linter; enables unstable rules and fixes.",
-          "scope": "window",
-          "type": "boolean",
-          "default": null
+          "default": null,
+          "markdownDescription": "Enable [preview mode](https://docs.astral.sh/ruff/settings/#lint_preview) for the linter; enables unstable rules and fixes.",
+          "scope": "resource",
+          "type": "boolean"
         },
         "ruff.lint.select": {
-          "markdownDescription": "Set rule codes to enable. Use `ALL` to enable all rules.",
+          "default": null,
+          "markdownDescription": "Set rule codes to enable. Use `ALL` to enable all rules. See [the documentation](https://docs.astral.sh/ruff/settings/#lint_select) for more details.",
+          "examples": [
+            ["E4", "E7", "E9", "F"]
+          ],
           "items": {
             "type": "string"
           },
-          "type": "array",
-          "default": null
+          "scope": "resource",
+          "type": "array"
         },
         "ruff.lint.extendSelect": {
+          "default": null,
           "markdownDescription": "Enable additional rule codes on top of existing configuration, instead of overriding it. Use `ALL` to enable all rules.",
           "items": {
             "type": "string"
           },
-          "type": "array",
-          "default": null
+          "scope": "resource",
+          "type": "array"
         },
         "ruff.lint.ignore": {
-          "markdownDescription": "Set rule codes to disable.",
+          "default": null,
+          "markdownDescription": "Set rule codes to disable. See [the documentation](https://docs.astral.sh/ruff/settings/#lint_ignore) for more details.",
           "items": {
             "type": "string"
           },
-          "type": "array",
-          "default": null
+          "scope": "resource",
+          "type": "array"
         },
         "ruff.run": {
           "default": "onType",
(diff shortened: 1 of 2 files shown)


MichaReiser:
Could you please share a print screen how the settings are rendered in VS code? 

What happens if you add settings to `ruff.lint.select` and then remove all of them again. Does it leave an empty list or does it reset the value to `null`?

src/common/settings.ts
```diff
@@ -127,6 +149,19 @@ function getGlobalValue<T>(config: WorkspaceConfiguration, key: string, defaultV
...
+  return inspect?.globalValue;
+}
+
+function getGlobalLineLength(config: WorkspaceConfiguration, key: string): number | undefined {
+  let lineLength = config.inspect<number>(key)?.globalValue;
+  if (lineLength && lineLength > 0 && lineLength <= 320) {
+    return lineLength;
+  }
+  return undefined;
+}
```
MichaReiser, line 163: Nit. We could avoid the repetion by passing the line-length value instead

```suggestion
function toLineLength(lineLength: number | undefined): number | undefined {
  if (lineLength && lineLength > 0 && lineLength <= 320) {
    return lineLength;
  }
  return undefined;
}
```

It can then be used in combination with `getOptionalGlobalValue`

src/common/settings.ts
```diff
@@ -127,6 +149,19 @@ function getGlobalValue<T>(config: WorkspaceConfiguration, key: string, defaultV
   return inspect?.globalValue ?? inspect?.defaultValue ?? defaultValue;
 }
 
+function getOptionalGlobalValue<T>(config: WorkspaceConfiguration, key: string): T | undefined {
```
MichaReiser, line 152: Nit: I prefer to use `null` instead of `undefined` for "absent" values because it is explicit that the value is not set whereas a value can be `undefined` if it is not set or not initializsed.

package.json
```diff
@@ -89,6 +89,36 @@
           "scope": "resource",
           "type": "array"
         },
+        "ruff.lint.preview": {
+          "markdownDescription": "Enable preview mode for the linter; enables unstable rules and fixes.",
+          "scope": "window",
```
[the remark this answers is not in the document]
snowsignal: I might change their scope to `resource`, though, this setting included.

src/common/settings.ts
snowsignal: `config.inspect`, the method from the VS Code API that we use to fetch configuration, uses `undefined` to represent absent values, which is why I think it makes more sense to also return `undefined` here.

Screenshot 2024-04-22 at 2 37 36 PM

snowsignal:
@MichaReiser 

> Could you please share a print screen how the settings are rendered in VS code?

It's hard to get a picture of all the new settings, since they're split up across the configuration. Here's what a few of them look like (`Lint > Run` is not a new setting):
Screenshot 2024-04-22 at 9 09 19 PM

> What happens if you add settings to ruff.lint.select and then remove all of them again. Does it leave an empty list or does it reset the value to null?

It leaves an empty list. VS Code indicates that the value is set by a blue highlight on it:
Screenshot 2024-04-22 at 9 11 10 PM

To make it `null` again, the user has to reset the setting from the dropdown menu. Which, to be fair, isn't particularly intuitive.

MichaReiser:
Thanks for the screenshot. Can we influence the ordering of the settings? E.g., I'm surprised that `lint.select`, `lint.extend-select`, and `lint.ignore` aren't next to each other (when you probably want to configure them together).

 > To make it null again, the user has to reset the setting from the dropdown menu. Which, to be fair, isn't particularly intuitive.

Yeah, that's not ideal. Maybe the documentation should mention that they either need to reset the configuration or manually edit the JSON. However, I think that's fine for now. We can still change the field type to `array | null` in the future. 

Having to click *Add Item* for each rule is a bit annoying but I guess users can navigate to the JSON editor themselves. So I think that UX is good.

MichaReiser (approved):

snowsignal:
> Thanks for the screenshot. Can we influence the ordering of the settings? E.g., I'm surprised that lint.select, lint.extend-select, and lint.ignore aren't next to each other (when you probably want to configure them together).

I thought this wasn't possible but it looks like this was recently added as a feature. 😄 I'll add ordering in a follow-up.

Issue #425, opened 2024-03-13
VSCode setting should be overriden by discovered toml


celynw:
Currently, extension settings override those which would be picked up in a `pyproject.toml`/`ruff.toml`.
This was highlighted in #3:

> Yeah, that's correct -- the `pyproject.toml` in the project path _doesn't_ override the settings from the VS Code extension. Which is not ideal, for the reasons you described.
> 
> The best way to solve this _today_ is to use a user-specific `pyproject.toml` rather than going through the VS Code extension. For example, if you're on macOS, you can create a `pyproject.toml` at `~/Library/Application\ Support/ruff/pyproject.toml`. Then, when linting, if Ruff can't find a `pyproject.toml` in the current file path, it will use _that_ as the default settings.

This is still the case (in my testing), but I'm surprised not to see another issue covering this (happy to be corrected!)
I think there should at least be an issue to track this one.

## Environment

- Ruff extension version `v2024.16.0` (`ruff==0.3.1`)
- VSCode version `1.87.2`
- OS: `Ubuntu 22.04.4`

Tested simply with `line-length = 120` in `ruff.toml`, and `--line-length=120` in VSCode setting `ruff.lint.args`

MichaReiser:
I think this needs some design work on the intended behavior (CC: @snowsignal). To me, the VS code settings taking precedence over the project settings seem reasonable, or at least I could see where the opposite behavior is the intended behavior.  It also opens the question if this should apply to all options or only some (e.g. what if I configure an additional configuration?) But I agree that what you're asking for makes sense, too; ultimately, the experience between IDE and CLI should be consistent, and that is only guaranteed by giving the project. to a higher priority.

celynw:
My use case is that I have synchronised VSCode settings everywhere. I'd like to be able to open different projects which each contain a `pyproject.toml` file. I either have to keep my 'master' synchronised Ruff VSCode settings empty, or I have to create a workspace settings file (under `.vscode/settings.json`) explicitly setting those settings to be empty, overriding my 'master' synced settings and allowing the toml file to do something.

But yes, I do see why one might expect the VSCode setting to take precedence instead!

snowsignal:
I'd support having a flag in the extension that lets discovered configuration take precedence. We'll look into solving this as we work on configuration for the upcoming VSCode extension based on our new Rust-based LSP.

dhruvmanila:
[...]

PR #457, opened 2024-04-22
Support prioritizing discovered TOML settings over editor settings

snowsignal:
## Summary

Fixes https://github.com/astral-sh/ruff-vscode/issues/425 (though only in the pre-release version, since this is a feature for the new LSP server)

This is a follow-up to https://github.com/astral-sh/ruff-vscode/pull/456.

A new setting has been introduced, `Prioritize File Configuration`, which tells the extension to prioritize settings from discovered TOML files over extension settings. Extension settings will still be used when a set field in the extension is not set in the file configuration.

## Test Plan

Refer to https://github.com/astral-sh/ruff/pull/11086 for a test plan.

### DIFF
--- a/package.json
+++ b/package.json
@@ -309,6 +309,12 @@
           "scope": "resource",
           "type": ["integer", "null"]
         },
+        "ruff.prioritizeFileConfiguration": {
+          "markdownDescription": "Whether to prioritize file configuration over configuration set in extension settings.",
+          "scope": "window",
+          "type": "boolean",
+          "default": false
+        },
         "ruff.enableExperimentalFormatter": {
           "default": false,
           "markdownDescription": "Controls whether Ruff registers as capable of code formatting.",
--- a/package.json
+++ b/package.json
@@ -309,11 +309,19 @@
           "scope": "resource",
           "type": ["integer", "null"]
         },
-        "ruff.prioritizeFileConfiguration": {
-          "markdownDescription": "Whether to prioritize file configuration over configuration set in extension settings.",
-          "scope": "window",
-          "type": "boolean",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -301,6 +301,12 @@
           "type": ["integer", "null"],
           "default": null
         },
+        "ruff.prioritizeFileConfiguration": {
+          "markdownDescription": "Whether to prioritize file configuration over configuration set in extension settings.",
+          "scope": "window",
+          "type": "boolean",
+          "default": false
+        },
```
MichaReiser, line 309: I wonder if this should be a dropdown instead. I could see multiple options

* Default: use the editor settings only if the project doesn't have a configuration file
* Override: Use the project settings but allow the editor settings to override individual settings
* Fallback: use the editor settings as fallbacks when the corresponding keys aren't set in the settings file.

I'm not very fond of the names and we don't need to implement all three variants today, but I think adding a dropdown gives us more flexibility in the future.

package.json
```diff
@@ -301,6 +301,12 @@
           "type": ["integer", "null"],
           "default": null
         },
+        "ruff.prioritizeFileConfiguration": {
+          "markdownDescription": "Whether to prioritize file configuration over configuration set in extension settings.",
```
MichaReiser, line 305: I don't think I would understand what that means. We may need to give an example here.

MichaReiser (approved):

package.json
snowsignal: I agree that a dropdown might be a better choice here. I'll update the `ruff` PR to reflect this as well.

PR #466, opened 2024-05-20
Integrate pre-release features into stable

snowsignal:
Note: this does not bump the version or create a new changelog entry. The goal of this PR is to integrate the changes made in the pre-release branch into the stable version for an upcoming release.


charliermarsh:
Why "beta server" over "experimental server" terminology?

snowsignal:
@charliermarsh I just thought the terminology could use a refresh now that the server has made significant improvements (and will soon become the new default).

charliermarsh:
@snowsignal - Changing the terminology seems reasonable, but I'd suggest picking something that's separate from the stability milestone. It kind of paints us into a corner. For example, what if we make the server opt-out once it stabilizes, but leave the setting in-place to facilitate that opt-out? We'll no longer be in the beta, but the setting would still be called `betaServer`... Or we'd have to rename it, which will require deprecation, etc.

What about `nativeServer` or something inherently related to the server, rather than the stability of it?

snowsignal:
@charliermarsh I like `nativeServer`!

charliermarsh (commented):
Overall LGTM, with the exception of changing to `nativeServer` terminology.

PR #468, opened 2024-05-22
Tweak the native server references

charliermarsh:


README.md
```diff
@@ -284,16 +286,34 @@ using the `ruff.lint.args` and `ruff.format.args` options in `settings.json`:
 Finally, to use a common Ruff configuration across all projects, consider creating a user-specific
 `pyproject.toml` or `ruff.toml` file as described in the [FAQ](https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings).
 
-### Using the Native Server
+### Enabling the Rust-based language server
```
dhruvmanila, line 289: Should we update the settings table to reflect the new ones? Or, maybe we can keep that as is and include a new table for all the supported settings with the new server.

charliermarsh: Good call. I will do that in a separate PR.

PR #469, opened 2024-05-22
Add supported settings to README

charliermarsh:


charliermarsh:
\cc @snowsignal @dhruvmanila

PR #487, opened 2024-06-04
Enable fix/formatting commands for Python and Jupyter Notebook files only

snowsignal:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/11744.

The extension would allow commands like `ruff.applyAutofix` to run on any file, even non-Python files. This PR adds a command palette filter so that these commands only appear in the palette when working with Python or Jupyter Notebook files.

## Test Plan

Ensure that commands like `Ruff: Fix all auto-fixable problems` still appear in the command palette (`Ctrl/Cmd + Shift + P`) for files with the following extensions:
* `*.py`
* `*.pyi`
* `*.ipynb`

Then, ensure that these commands do not appear for other kinds of files (for example: `*.toml` or `*.json`). `Ruff: Restart server` should still be available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -370,7 +370,23 @@
         "category": "Ruff",
         "command": "ruff.restart"
       }
-    ]
+    ],
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "ruff.executeAutofix",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
+        },
+        {
+          "command": "ruff.executeFormat",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
+        },
+        {
+          "command": "ruff.executeOrganizeImports",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
+        }
+      ]
+    }
   },
   "dependencies": {
     "fs-extra": "^11.1.1",


dhruvmanila (approved):
It might be worth looking at how ESLint does it as it doesn't raise any error if we run the "ESLint: fix all" command in settings.json.

package.json
```diff
@@ -370,7 +370,23 @@
...
+          "command": "ruff.executeAutofix",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
+        },
+        {
+          "command": "ruff.executeFormat",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
+        },
+        {
+          "command": "ruff.executeOrganizeImports",
+          "when": "resourceExtname == .py || resourceExtname == .pyi || resourceExtname == .ipynb"
```
MichaReiser, line 386: Hmm, I don't think we can do that. You can configure ruff to lint files with arbitrary extensions using `extend-include`. These files would no longer be linted. It also means that auto fix would no longer run on any new/untitled files because they don't have an extension yet (they only have an associated language that we should probably use in the LSP).

MichaReiser (changes requested):
I don't think this works for untitled documents or files that use custom extensions but are associated with the python language.

package.json
snowsignal: You're right, this wouldn't work with arbitrary file extensions.

I might try using @dhruvmanila's suggestion and just handle the `snapshot not available` error more gracefully in `ruff server`.

PR #114, opened 2023-01-19
Expose the ruff.trace.server setting for tracing ruff-lsp activity

yaegassy:
## Description

Add the `*.trace.server` setting to the configuration option, so that configuration completion also works in `settings.json`.

**VSCode Reference**:

- <https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#logging-support-for-language-server>

**Examples of other extensions**:

- <https://github.com/redhat-developer/vscode-yaml/blob/main/package.json#L104-L113>
- and more...

## DEMO (mp4)

https://user-images.githubusercontent.com/188642/213360027-b48bcb32-3b49-4f5c-8301-135b3571e23d.mp4

PR #489, opened 2024-06-05
Remove the server trace level override that was based on the extension log level

### LINKED ISSUE #488: The extension does not respect the `ruff.trace.server` setting on startup
According to the [VS Code extension documentation](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#logging-support-for-language-server), `[langId].trace.server` is used to set the trace value for both the client and the server.

However, in the Ruff extension, this value is ignored on startup, because it is immediately overridden when we set the trace value from `getLSClientTraceLevel`. This function determines the trace level from the output channel's log level. What this means is that by default, Ruff's _actual_ trace level is almost always `"messages"` instead of `"off"`, which is the default value of `ruff.trace.server`. This is confusing and counterintuitive - if you actually want to turn off tracing messages in Ruff, you'll have to change `ruff.trace.server` _twice_ from the extension settings window (once to a different value from the one you originally set, and then back to the intended value). And this will get clobbered when the server restarts, so you'll have to do this a lot.

I propose that we remove `getLSClientTraceLevel` altogether and use `ruff.trace.server` as a single configuration point for tracing.

snowsignal:
## Summary

Fixes #488.

We no longer override the trace level on server restarts. This means that the trace level will always be set to the value of `ruff.trace.server`.

## Test Plan

With `ruff.trace.server` set to `"off"`, you should no longer see logs in the output marked as `[Trace - ...]`.

Before:
Screenshot 2024-06-05 at 11 05 41 AM

After:
Screenshot 2024-06-05 at 11 06 56 AM

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -24,7 +24,7 @@ import {
   ISettings,
 } from "./settings";
 import { updateStatus } from "./status";
-import { getLSClientTraceLevel, getProjectRoot } from "./utilities";
+import { getProjectRoot } from "./utilities";
 import { isVirtualWorkspace } from "./vscodeapi";
 
 export type IInitOptions = {
@@ -201,7 +201,5 @@ export async function restartServer(
     return undefined;
   }
 
-  const level = getLSClientTraceLevel(outputChannel.logLevel, env.logLevel);
-  await newLSClient.setTrace(level);
   return newLSClient;
 }
--- a/src/common/utilities.ts
+++ b/src/common/utilities.ts
@@ -22,19 +22,6 @@ function logLevelToTrace(logLevel: LogLevel): Trace {
   }
 }
 
-export function getLSClientTraceLevel(channelLogLevel: LogLevel, globalLogLevel: LogLevel): Trace {
-  if (channelLogLevel === LogLevel.Off) {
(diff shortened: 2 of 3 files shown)


dhruvmanila (approved):

MichaReiser:
Do you know what the motivation was for using `getLSClientTraceLevel` in the first place? Do we still need to support `getLSClientTraceLevel` for the python LSP?

Personal preference: I like to include screenshots when making changes that change the visual output. The screenshot alone can communicate a lot of context to the reviewer. It's like, ahhh, you changed that part and it now looks like this. I like it!

snowsignal:
@MichaReiser 

> Do you know what the motivation was for using getLSClientTraceLevel in the first place?

We took it from the [Python tools extension template](https://github.com/microsoft/vscode-python-tools-extension-template) when we started building the extension. The template doesn't have a `[langId].trace.server` setting - this was added later, in #114, and this was never removed. That's why we've had two competing setters for the trace level.

> Do we still need to support `getLSClientTraceLevel` for the python LSP?

No, it also uses the `[langId].trace.server` setting.

> Personal preference: I like to include screenshots when making changes that change the visual output.

I've added screenshots 👍

PR #495, opened 2024-06-10
Introduce `ruff.printDebugInformation` command

snowsignal:
## Summary

Introduces a command to print debug information. It should look like this in the Output window:

Screenshot 2024-06-10 at 12 21 07 PM

## Test Plan

See https://github.com/astral-sh/ruff/pull/11831 for testing information.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -191,6 +191,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
       });
     }),
     registerCommand(`${serverId}.executeFormat`, async () => {
+      // let configuration = getConfiguration(serverId) as ISettings;
+
       if (!lsClient) {
         return;
       }
@@ -240,8 +242,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         );
       });
     }),
-    registerCommand(`${serverId}.printDebugInformation`, async () => {
-      if (!lsClient) {
+    registerCommand(`${serverId}.debugInformation`, async () => {
+      let configuration = getConfiguration(serverId) as unknown as ISettings;
+      if (!lsClient || !configuration.nativeServer) {
         return;
       }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -240,6 +240,19 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         );
       });
     }),
+    registerCommand(`${serverId}.printDebugInformation`, async () => {
+      if (!lsClient) {
+        return;
+      }
+
+      const params = {
+        command: `${serverId}.printDebugInformation`,
+      };
+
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):

src/extension.ts
```diff
@@ -240,6 +240,19 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         );
       });
     }),
+    registerCommand(`${serverId}.printDebugInformation`, async () => {
+      if (!lsClient) {
```
MichaReiser, line 244: Can we only register the command if this is the new LSP server because I suspect it isn't supported for `ruff-lsp`?

dhruvmanila: Yeah, I agree. `ruff-lsp` doesn't have the `ruff.printDebugInformation` command in its capabilities so it'll throw an error.

dhruvmanila: Confirmed that it'll raise an exception:
```
Failed to handle request 2 workspace/executeCommand ExecuteCommandParams(command='ruff.printDebugInformation', arguments=[{'uri': 'file:///Users/dhruv/playground/ruff/src/TCH002.py'}], work_done_token=None)
  File "/Users/dhruv/.local/state/pipx/venvs/ruff-lsp/lib/python3.12/site-packages/pygls/protocol/json_rpc.py", line 264, in _handle_request
    handler(params, msg_id)
  File "/Users/dhruv/.local/state/pipx/venvs/ruff-lsp/lib/python3.12/site-packages/pygls/protocol/lsp_meta.py", line 17, in decorator
    ret_val = base_func(self, *args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dhruv/.local/state/pipx/venvs/ruff-lsp/lib/python3.12/site-packages/pygls/protocol/language_server.py", line 320, in lsp_workspace__execute_command
    cmd_handler = self.fm.commands[params.command]
                  ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'ruff.printDebugInformation'
```

snowsignal:
I've updated the command handler to ignore the command when we're using `ruff-lsp` (and I've added `(name server only)` to the command to make this more clear).

src/extension.ts
```diff
@@ -191,6 +191,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
       });
     }),
     registerCommand(`${serverId}.executeFormat`, async () => {
+      // let configuration = getConfiguration(serverId) as ISettings;
+
```
dhruvmanila, line 195: Is this a leftover? ;)

snowsignal: Good catch, I'll remove that

PR #502, opened 2024-06-25
Migrate from `pip-tools` to `uv`

charliermarsh:
## Summary

Biggest benefit here is that we no longer need to run resolution on Python 3.7 -- we can just pass `--python-version` to uv.


justfile
```diff
@@ -1,31 +1,31 @@
...
-  pip install --no-deps -r ./requirements-dev.txt
+  uv pip sync ./requirements-dev.txt
   npm ci
 
 test: setup
   python -m unittest
 
 check:
-  ruff ./bundled/tool ./build ./tests
+  ruff check ./bundled/tool ./build ./tests
```
charliermarsh, line 20: This is about to be removed so...

PR #504, opened 2024-06-27
Add `showSyntaxErrors` extension setting

dhruvmanila:
## Summary

This PR adds a `ruff.showSyntaxErrors` config option for https://github.com/astral-sh/ruff/pull/12059.

## Test Plan

Refer to the test plan in https://github.com/astral-sh/ruff/pull/12059.

### DIFF
--- a/package.json
+++ b/package.json
@@ -271,7 +271,7 @@
         },
         "ruff.showSyntaxErrors": {
           "default": true,
-          "markdownDescription": "Controls whether Ruff should show syntax error diagnostics.",
+          "markdownDescription": "Whether to show syntax error diagnostics.",
           "scope": "window",
           "type": "boolean"
         },
--- a/package.json
+++ b/package.json
@@ -269,6 +269,12 @@
           "additionalProperties": false,
           "markdownDescription": "Whether to display Quick Fix actions to disable rules via `noqa` suppression comments."
         },
+        "ruff.showSyntaxErrors": {
+          "default": true,
+          "markdownDescription": "Controls whether Ruff should show syntax error diagnostics.",
+          "scope": "window",
+          "type": "boolean"
+        },
         "ruff.ignoreStandardLibrary": {
           "default": true,
           "markdownDescription": "Whether to ignore files that are inferred to be part of the Python standard library.",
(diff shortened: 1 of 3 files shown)


snowsignal (approved):

package.json
```diff
@@ -269,6 +269,12 @@
           "additionalProperties": false,
           "markdownDescription": "Whether to display Quick Fix actions to disable rules via `noqa` suppression comments."
         },
+        "ruff.showSyntaxErrors": {
+          "default": true,
+          "markdownDescription": "Controls whether Ruff should show syntax error diagnostics.",
```
MichaReiser, line 274: To align with other descriptions
```suggestion
          "markdownDescription": "Whether to show syntax error diagnostics.",
```

package.json
```diff
@@ -269,6 +269,12 @@
           "additionalProperties": false,
           "markdownDescription": "Whether to display Quick Fix actions to disable rules via `noqa` suppression comments."
         },
+        "ruff.showSyntaxErrors": {
+          "default": true,
```
MichaReiser, line 273: What was the default before. Did ruff always filter out syntax error diagnostics? I'm mainly asking to understand if this is a "breaking" change in the sense that the default changes

MichaReiser (approved):
LGTM.

We should add the new setting to the README and maybe mention that the setting is only supported with Ruff 0.5.

package.json
dhruvmanila: No, Ruff would always show syntax error diagnostics.

MichaReiser: Good :) Unless someone used `ignore=E999` but that "breaking" change is already handled by Ruff's changelog

dhruvmanila: Yeah, I assume you mean `ruff.ignore` server setting. Ruff would emit a warning in that case but it won't be displayed to the user, it'll be in the logs:
```
2024-06-27 12:34:00.546 [info] warning: Rule `E999` is deprecated and will be removed in a future release. Syntax errors will always be shown regardless of whether this rule is selected or not.
```
But, this won't work for `ignore` because it doesn't work in Ruff itself.

PR #505, opened 2024-06-27
Support automatic `ruff server` enablement

snowsignal:
## Summary

The native server is now automatically enabled when using Ruff `v0.4.5` or later. The extension can also be configured to always use the native server (or never use it) via the new `languageServer` setting.

## Test Plan

1. Uninstall any local Ruff executables from your system.
2. Launch the extension. It should be using `ruff server` by default, since the extension uses a compatible bundled version (you can distinguish whether or not the native server is running by the log output)
3. Switch the `languageServer` setting to `legacy`. The extension should now be using `ruff-lsp`.
4. Switch the `languageServer` setting back to `automatic`.
5. Install Ruff version `v0.4.5`. Confirm that the extension uses this executable, and that it's using the native server.
6. Install a Ruff version before `v0.4.5`. Confirm that the extension uses this executable, and that it's using `ruff-lsp`.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -14,6 +14,7 @@ import {
   RUFF_SERVER_CMD,
   SERVER_SCRIPT_PATH,
   EXPERIMENTAL_SERVER_SCRIPT_PATH,
+  RUFF_VERSION_ARGS,
 } from "./constants";
 import { traceError, traceInfo, traceVerbose } from "./log/logging";
 import { getDebuggerPath } from "./python";
@@ -26,6 +27,10 @@ import {
 import { updateStatus } from "./status";
 import { getProjectRoot } from "./utilities";
 import { isVirtualWorkspace } from "./vscodeapi";
+import { promisify } from "util";
+import { exec } from "child_process";
+
+const MIN_NATIVE_VERSION = [0, 4, 5];
 
 export type IInitOptions = {
   settings: ISettings[];
@@ -87,7 +92,7 @@ async function createNativeServer(
   return new LanguageClient(serverId, serverName, serverOptions, clientOptions);
 }
 
-async function createServer(
+async function createLegacyServer(
   settings: ISettings,
   serverId: string,
   serverName: string,
@@ -142,6 +147,26 @@ async function createServer(
   return new LanguageClient(serverId, serverName, serverOptions, clientOptions);
 }
 
+async function useNative(settings: ISettings): Promise<boolean> {
+  switch (settings.languageServer) {
+    case "automatic":
+      let version = await getRuffVersion(settings);
+      for (let i = 0; i < 3; i++) {
+        if (version[i] > MIN_NATIVE_VERSION[i]) {
+          return true;
+        }
+        if (version[i] < MIN_NATIVE_VERSION[i]) {
+          return false;
+        }
+      }
+      return true;
+    case "native":
+      return true;
+    case "legacy":
+      return false;
+  }
+}
+
 let _disposables: Disposable[] = [];
 export async function restartServer(
   serverId: string,
@@ -165,13 +190,13 @@ export async function restartServer(
   const globalSettings = await getGlobalSettings(serverId);
 
   let newLSClient;
-  if (workspaceSettings.nativeServer || globalSettings.nativeServer) {
+  if (await useNative(workspaceSettings)) {
     newLSClient = await createNativeServer(workspaceSettings, serverId, serverName, outputChannel, {
       settings: extensionSettings,
       globalSettings: globalSettings,
     });
   } else {
-    newLSClient = await createServer(workspaceSettings, serverId, serverName, outputChannel, {
+    newLSClient = await createLegacyServer(workspaceSettings, serverId, serverName, outputChannel, {
       settings: extensionSettings,
       globalSettings: globalSettings,
     });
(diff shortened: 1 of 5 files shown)


src/common/server.ts
```diff
@@ -142,6 +147,26 @@ async function createServer(
...
+  switch (settings.languageServer) {
+    case "automatic":
+      let version = await getRuffVersion(settings);
+      for (let i = 0; i < 3; i++) {
+        if (version[i] > MIN_NATIVE_VERSION[i]) {
+          return true;
+        }
+        if (version[i] < MIN_NATIVE_VERSION[i]) {
+          return false;
+        }
```
MichaReiser, line 160: I think we should also consider what settings they are using and **not** enable the new server when they use any incompatible settings.

src/common/server.ts
```diff
@@ -203,3 +228,29 @@ export async function restartServer(
 
   return newLSClient;
 }
+
+async function getRuffVersion(settings: ISettings): Promise<[number, number, number]> {
+  let command;
+  if (settings.path[0]) {
```
MichaReiser, line 234: Nit
```suggestion
  if (settings.path[0] != null) {
```

src/common/server.ts
```diff
@@ -203,3 +228,29 @@ export async function restartServer(
...
+    command = `${settings.path[0]} ${RUFF_VERSION_ARGS.join(" ")}`;
+  } else {
+    command = `${
+      settings.interpreter[0]
```
MichaReiser, line 238: What's the reason we use `settings.interpreter` here directly over `getInterpreterFromSetting`?

src/common/server.ts
```diff
@@ -203,3 +228,29 @@ export async function restartServer(
...
+    } ${EXPERIMENTAL_SERVER_SCRIPT_PATH} ${RUFF_VERSION_ARGS.join(" ")}`;
+  }
+  traceInfo(`Checking Ruff's version with command: ${command}`);
+  const execCommand = promisify(exec);
+  const { stdout } = await execCommand(command, {
+    cwd: settings.cwd,
+  });
```
MichaReiser, line 245: What happens if executing the command fails? For example, because the ruff path is incorrect?

src/common/server.ts
```diff
@@ -203,3 +228,29 @@ export async function restartServer(
...
+  const versionRegex = /ruff (\d+).(\d+).(\d+)/;
+  const matches = stdout.match(versionRegex) ?? [];
+  let major = Number(matches[1] ?? "X");
+  let minor = Number(matches[2] ?? "X");
+  let patch = Number(matches[3] ?? "X");
```
MichaReiser, line 250: I think we want to constrain this to integers only
```suggestion
  let major = parseInt(matches[1] ?? "X", 10);
  let minor = parseInt(matches[2] ?? "X", 10);
  let patch = parseInt(matches[3] ?? "X", 10);
```

src/common/server.ts
```diff
@@ -142,6 +147,26 @@ async function createServer(
   return new LanguageClient(serverId, serverName, serverOptions, clientOptions);
 }
 
+async function useNative(settings: ISettings): Promise<boolean> {
+  switch (settings.languageServer) {
```
MichaReiser, line 151: Let's add a log showing which server the extension is using and why.

src/common/server.ts
MichaReiser: I could see that work too. But I think we at least need warnings in that case that inform users that some of their settings were ignored.

MichaReiser: Considering how we deprecate other features, I think we should actually consider settings here as well, at least for now. Our normal deprecation procedure is:

* Deprecate the feature, warn about usage, but it remains functioning
* Remove the feature, but make it a hard error
* Remove the feature entirely

In that sense, I would expect that the extension keeps functioning without any changes if I use "old" settings but that it would show me a deprecation warning.

src/common/server.ts
snowsignal: If that happens, `stdout` will either output nothing or an invalid string, in which case the version will fail to parse and return as `[0, 0, 0]`

snowsignal: (which means that we will never select the native server)

src/common/server.ts
snowsignal: This is how `createNativeServer` and `createLegacyServer` get the interpreter path - the settings passed in are initialized from `getWorkspaceSettings`, which itself calls `getInterpreterFromSetting` to set `settings.interpreter`.

src/common/server.ts
MichaReiser: What I understand what happens if e.g. the path to the ruff binary is invalid is that `execCommand` fails (the promise gets rejected). Is there any upstream code that catches the exception? Is defaulting to the `ruff-lsp` the right fallback in that case or should we instead show a message saying that the path is invalid?

snowsignal: > Is there any upstream code that catches the exception?

Not that I'm aware of, but this was also the current behavior of the extension before this change - if the path is invalid, the server crashes. I'd definitely be in favor of handling this more gracefully, but I'd prefer doing that in a follow-up.

MichaReiser: I'm fine handling this in a follow up PR. But can we make sure that the output is about as good as for the current LSP?

```
2024-06-28 14:18:37.913 [info] [Error - 2:18:37 PM] Ruff client: couldn't create connection to server.
2024-06-28 14:18:37.913 [info] Launching server using command /home/micha/astral/ruff/target/debug/ruff2 failed. Error: spawn /home/micha/astral/ruff/target/debug/ruff2 ENOENT
2024-06-28 14:18:37.913 [error] Server: Start failed: Launching server using command /home/micha/astral/ruff/target/debug/ruff2 failed. Error: spawn /home/micha/astral/ruff/target/debug/ruff2 ENOENT
2024-06-28 14:18:37.913 [info]   22.461738414s  INFO ruff:main ruff_server::server::connection: Exit notification received. Server shutting down...
```

snowsignal: Okay, I've tested this and it actually seems like we do just fall-back to `ruff-lsp`. I think that's fine for now, we can add the warning in a follow-up.

src/common/settings.ts
```diff
@@ -39,6 +41,7 @@ type Format = {
 
 export interface ISettings {
   nativeServer: boolean;
+  languageServer: LanguageServer;
```
T-256, line 44: Do we have plan to introduce new type of server in future? otherwise, I'd recommend using one setting and avoid making more scenarios that settings maybe not worked as users expected when modified together.

How about `nativeServer` setting with values of `auto | on | off` which is closer to alternative vscode's settings? we can keep it backward compatible while introducing new values.
I think when `nativeServer` chosen at https://github.com/astral-sh/ruff-vscode/pull/466#issuecomment-2122463407, we also realized that it could have values expanded in future.

dhruvmanila: > How about `nativeServer` setting with values of `auto | on | off` which is closer to alternative vscode's settings?

This sounds like a pretty reasonable option to me. We can keep the "boolean" type for backward compatibility and map `true`/`false` to `on`/`off` respectively.

snowsignal: Good idea @T-256! I'll rework this setting.

src/common/server.ts
dhruvmanila: Ok, I agree that we should look at the settings as well. Now, as the user explicitly opted into using the "auto" behavior, we should be providing warning.

dhruvmanila:
Closing in favor of https://github.com/astral-sh/ruff-vscode/pull/518

Issue #476, opened 2024-05-23
Better notification when used Ruff not compatible to extension.


T-256:
Notifications on using Ruff v0.3.0 with `ruff.nativeServer` settings enabled:

![image](https://github.com/astral-sh/ruff-vscode/assets/132141463/bc1e5f21-9ce5-44b0-9278-48f7e2a000d4)

### Output
```log
2024-05-23 13:26:39.572 [info] Server: Stop requested
2024-05-23 13:26:39.585 [info] Server run command: c:\Users\User\Desktop\test\.venv\Scripts\python.exe c:\Users\User\.vscode\extensions\charliermarsh.ruff-2024.22.0-win32-x64\bundled\tool\ruff_server.py server --preview
2024-05-23 13:26:39.585 [info] Server: Start requested.
2024-05-23 13:26:39.591 [info] 2024-05-23 13:26:39,580 INFO Shutting down the server
2024-05-23 13:26:39,580 INFO Closing the event loop.

2024-05-23 13:26:39.759 [info] Traceback (most recent call last):
  File "c:\Users\User\.vscode\extensions\charliermarsh.ruff-2024.22.0-win32-x64\bundled\tool\ruff_server.py", line 90, in <module>
    check_compatibility(ruff, RUFF_VERSION_REQUIREMENT)
  File "c:\Users\User\.vscode\extensions\charliermarsh.ruff-2024.22.0-win32-x64\bundled\tool\ruff_server.py", line 55, in check_compatibility
    raise RuntimeError(message)
RuntimeError: Ruff >=0.3.5 required, but found 0.3.0 at C:\Users\User\.rye\shims\ruff.EXE

2024-05-23 13:26:39.767 [info] [Error - 1:26:39 PM] Server initialization failed.
2024-05-23 13:26:39.767 [info]   Message: Pending response rejected since connection got disposed
  Code: -32097 
2024-05-23 13:26:39.767 [info] [Info  - 1:26:39 PM] Connection to server got closed. Server will restart.
2024-05-23 13:26:39.767 [info] true
2024-05-23 13:26:39.768 [info] [Error - 1:26:39 PM] Ruff client: couldn't create connection to server.
2024-05-23 13:26:39.768 [info]   Message: Pending response rejected since connection got disposed
  Code: -32097 
2024-05-23 13:26:39.914 [info] Traceback (most recent call last):
[...]

Issue #508, opened 2024-06-29
Graceful notification for unsupported features.


T-256:
Since the default behavior of this extension is to use ruff binary from environment, sometimes users update the extension but still use outdated version of ruff binary in their environment.
So, when new feature available to the extension's settings/commands, it may throw multiple error notifications to the users about ruff crashing.
potential solution is specifying minimum required ruff version at extension side for new settings/commands. so, when that feature has been used, we can ignore it and show user a warning instead.

dhruvmanila:
Is this similar to https://github.com/astral-sh/ruff-vscode/issues/476?

PR #512, opened 2024-07-03
Better error message for unsupported Ruff version

dhruvmanila:
## Summary

This PR adds back the message to suggest the user that the current Ruff binary doesn't support the native server.

This was removed from the Python script in #511.

fixes: #476 

## Test Plan

Install an older Ruff version to test this out using `pipx install --force "ruff==0.3.3"`.

Screenshot 2024-07-03 at 14 50 41

And, we've another log message which displays the Ruff version similar to `ruff-lsp`:
```
2024-07-03 16:00:03.089 [info] Using the Ruff binary: /Users/dhruv/work/astral/ruff-lsp/.venv/bin/ruff
2024-07-03 16:00:03.094 [info] Found Ruff 0.4.10 at /Users/dhruv/work/astral/ruff-lsp/.venv/bin/ruff
2024-07-03 16:00:03.095 [info] Server run command: /Users/dhruv/work/astral/ruff-lsp/.venv/bin/ruff server --preview
```

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -51,6 +51,54 @@ function executeCommand(command: string): Promise<string> {
   });
 }
 
+type VersionInfo = {
+  major: number;
+  minor: number;
+  patch: number;
+};
+
+/**
+ * Convert a version object to a string.
+ */
+function versionToString(version: VersionInfo): string {
+  return `${version.major}.${version.minor}.${version.patch}`;
+}
+
+/**
+ * The minimum version of the Ruff executable that supports the native server.
+ */
+const MINIMUM_RUFF_SERVER_VERSION: VersionInfo = { major: 0, minor: 3, patch: 5 };
+
+/**
+ * Check if the given version of the Ruff executable supports the native server.
+ */
+function supportsNativeServer(version: VersionInfo): boolean {
+  if (version.major > MINIMUM_RUFF_SERVER_VERSION.major) {
+    return true;
+  }
+  if (version.major === MINIMUM_RUFF_SERVER_VERSION.major) {
+    if (version.minor > MINIMUM_RUFF_SERVER_VERSION.minor) {
+      return true;
+    }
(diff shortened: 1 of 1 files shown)


src/common/server.ts
```diff
@@ -48,6 +49,54 @@ function executeCommand(command: string): Promise<string> {
...
+ * Convert a version object to a string.
+ */
+function versionToString(version: VersionInfo): string {
+  return `${version.major}.${version.minor}.${version.patch}`;
+}
+
+/**
+ * The minimum version of the Ruff executable that supports the native server.
+ */
+const MINIMUM_RUFF_SERVER_VERSION: VersionInfo = { major: 0, minor: 3, patch: 5 };
```
dhruvmanila, line 68: This comes from https://github.com/astral-sh/ruff-vscode/pull/443 although the minimum supported version if actually `0.3.3` but I guess this is the version we moved away from using the bundled version.

T-256: As explained in https://github.com/astral-sh/ruff-vscode/issues/508 there are also commands/features that unsupported by old version of nativeServer and/or ruff-lsp. should we declare version constant for them? e.g. `ruff.printDebugInformation` which was added in https://github.com/astral-sh/ruff/pull/11831 for nativeServer

dhruvmanila: Good point. I intend to implement the warning part for the settings in https://github.com/astral-sh/ruff-vscode/pull/505.

> `ruff.printDebugInformation` which was added in [astral-sh/ruff#11831](https://github.com/astral-sh/ruff/pull/11831) for nativeServer

Regarding this command, I think we should only register a VS Code command only if it's available in the server capabilities (https://github.com/astral-sh/ruff/pull/11850). This way we don't need to check the Ruff version for at least the server capabilities.

dhruvmanila: I've opened #513 to keep track of it.

MichaReiser (approved):

charliermarsh (approved):

PR #516, opened 2024-07-05
Validate Python interpreter version for native server

dhruvmanila:
## Summary

I'm not sure why this check was removed looking back at https://github.com/astral-sh/ruff-vscode/pull/431, but this validation should be done for both native server and `ruff-lsp`. Both uses a Python script which requires Python 3.7 or later and `ruff` in general also supports 3.7 and later.

Without this change, if someone were to use Python 3.6, the extension would give the following error:
```
2024-07-05 09:26:21.378 [error] Error while trying to find the Ruff binary: Error: Command failed: /Users/dhruv/.pyenv/versions/3.6.15/bin/python /Users/dhruv/work/astral/ruff-vscode/bundled/tool/find_ruff_binary_path.py
  File "/Users/dhruv/work/astral/ruff-vscode/bundled/tool/find_ruff_binary_path.py", line 33, in <module>
    ruff_binary_path = find_ruff_binary_path()
  File "/Users/dhruv/work/astral/ruff-vscode/bundled/tool/find_ruff_binary_path.py", line 20, in find_ruff_binary_path
    elif sys.platform == "darwin" and sys._framework:
AttributeError: module 'sys' has no attribute '_framework'
```

## Test Plan

The above error still occurs and the fix for that is in a follow-up PR (https://github.com/astral-sh/ruff-vscode/pull/517).

Using the below settings:
```json
{
  "ruff.nativeServer":true,
  "ruff.interpreter": ["/Users/dhruv/.pyenv/versions/3.6.15/bin/python"]
}
```

The logs shows that the version isn't supported:
```
2024-07-05 09:27:38.086 [error] Python version 3.6 is not supported.
2024-07-05 09:27:38.086 [error] Selected python path: /Users/dhruv/.pyenv/versions/3.6.15/bin/python
2024-07-05 09:27:38.086 [error] Supported versions are 3.7 and above.
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -55,7 +55,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     }),
   );
 
-  const { enable, nativeServer } = getConfiguration(serverId) as unknown as ISettings;
+  const { enable } = getConfiguration(serverId) as unknown as ISettings;
   if (!enable) {
     traceLog(
       "Extension is disabled. To enable, change `ruff.enable` to `true` and restart VS Code.",
@@ -86,9 +86,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 
     restartInProgress = true;
 
-    if (nativeServer) {
-      traceVerbose("Using experimental server with bundled Ruff binary");
-
+    const interpreter = getInterpreterFromSetting(serverId);
+    if (
+      interpreter !== undefined &&
+      interpreter.length > 0 &&
+      checkVersion(await resolveInterpreter(interpreter))
+    ) {
+      traceVerbose(
+        `Using interpreter from ${serverInfo.module}.interpreter: ${interpreter.join(" ")}`,
+      );
       lsClient = await restartServer(serverId, serverName, outputChannel, lsClient);
 
       restartInProgress = false;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -86,9 +86,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 
     restartInProgress = true;
 
-    if (nativeServer) {
-      traceVerbose("Using experimental server with bundled Ruff binary");
-
+    const interpreter = getInterpreterFromSetting(serverId);
+    if (
+      interpreter &&
+      interpreter.length > 0 &&
```
MichaReiser, line 92:
```suggestion
      interpreter?.length > 0 &&
```

MichaReiser: or does typescript complain that the left isn't a number. In that case I would write `interpreter != null && interpreter.length > 0`. We should probably enable the lint rule to enforce `!=` instead of on truthiness checks.

MichaReiser (approved):

src/extension.ts
dhruvmanila: Yeah, the compiler complains. I'll change it to use explicit `null` checks but I can't find the lint rule which corresponds to that in https://www.typescriptlang.org/tsconfig/.

MichaReiser: I think it's https://typescript-eslint.io/rules/strict-boolean-expressions/

PR #519, opened 2024-07-08
Show server kind in the status bar

dhruvmanila:
## Summary

This PR updates the status item to show the chosen server in the status bar. The value would be either `Ruff (native)` or `Ruff (ruff-lsp)`.

## Test Plan

When `nativeServer` is `true`:

Screenshot 2024-07-08 at 15 50 46

When `nativeServer` is `false`:

Screenshot 2024-07-08 at 15 51 00


MichaReiser (approved):
This is a nice addition!

PR #527, opened 2024-07-12
Eagerly activate the Python extension for tests

dhruvmanila:


dhruvmanila:
I think we just need to wait until both extensions are completely activated and only then proceed.

PR #530, opened 2024-07-16
Add `logLevel` and `logFile` extension setting

dhruvmanila:
## Summary

This PR adds `ruff.logLevel` and `ruff.logFile` settings to the VS Code extension which corresponds to the same for the `ruff server`. Without these, there's no way to configure the log level and the destination for the log messages unlike other editors.

## Test Plan

Using the following config:
```json
{
  "ruff.logLevel": "debug",
  "ruff.logFile": "~/playground/ruff/server.logs",
  "ruff.trace.server": "messages"
}
```

I confirmed that it outputs the logs messages at the specified location.

### DIFF
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -16,6 +16,8 @@ type ConfigPreference = "editorFirst" | "filesystemFirst" | "editorOnly";
 
 type NativeServer = boolean | "on" | "off" | "auto";
 
+type LogLevel = "error" | "warn" | "info" | "debug" | "trace";
+
 type CodeAction = {
   disableRuleComment?: {
     enable?: boolean;
@@ -60,6 +62,8 @@ export interface ISettings {
   lineLength?: number;
   configurationPreference?: ConfigPreference;
   showSyntaxErrors: boolean;
+  logLevel?: LogLevel;
+  logFile?: string;
 }
 
 export function getExtensionSettings(namespace: string): Promise<ISettings[]> {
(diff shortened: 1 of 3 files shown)


MichaReiser (commented):
What's the relationship of the new `logLevel` setting to the tracing setting?

dhruvmanila:
> What's the relationship of the new `logLevel` setting to the tracing setting?

The `ruff.server.trace` value is either `off | messages | verbose` where
* `off` doesn't emit any log messages
* `messages` will only emit the type of request and response along with Ruff server's log messages but only at `INFO` or above
* `verbose` will emit the request and response parameters as well

Ruff server also emits trace messages at different levels, specifically `DEBUG` level but there's no way to update that from VS Code. This is different in other editors where a user can update the `logLevel` server setting.

Only updating the `logLevel` doesn't make any difference, it should be combined with updating the `ruff.server.trace` to `messages` or `verbose` for Ruff to emit the messages at the specified `logLevel`.

src/common/settings.ts
```diff
@@ -126,6 +130,11 @@ export async function getWorkspaceSettings(
     configuration = resolveVariables(configuration, workspace);
   }
 
+  let logFile = config.get<string>("logFile");
+  if (logFile !== undefined) {
+    logFile = resolveVariables(logFile, workspace);
+  }
+
```
dhruvmanila, line 137: Actually, this is incorrect. It seems that the `logLevel` and `logFile` settings in the server code is only taken from the global settings i.e., they aren't resolve per workspace which makes sense. So, we can't allow VS Code variables in here. I'll revert this.

MichaReiser (approved):

Issue #520, opened 2024-07-08
Update Ruff version when the server is stabilized


dhruvmanila:
This should be updated to reflect the version of the stable `ruff server` before updating the extension:

https://github.com/astral-sh/ruff-vscode/blob/4a54d4d97e65eb49c358a8c6320c8ff482c6d590/src/common/version.ts#L49-L52

PR #531, opened 2024-07-17
Update to reflect the stabilized `ruff server`

dhruvmanila:
## Summary

This PR updates the README to reflect the stabilized `ruff server`. It also changes the minimum required version for the usage of `ruff server` in `nativeServer: auto` to be `0.5.3`.

For the README, the settings that are just a _pass through_ to the native server are removed and instead we point the users to the online settings reference. Now, the settings table only contains the ones which are specific to VS Code extension.

For review, here's the rendered README: https://github.com/astral-sh/ruff-vscode/blob/dhruv/readme/README.md

resolves: #520

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Refer to the [Ruff Language Server documentation](https://docs.astral.sh/ruff/se
 settings available in the extension.
 
 The extension provides the following settings in addition to those available in the Ruff Language Server that can
-be used to configure the extension. The detailed documentation for these settings can be found in the settings
+be used to configure the extension. Detailed documentation for these settings can be found in the settings
 editor in Visual Studio Code.
 
 | Settings                    | Default           | Description                                                                                                                                                                                 |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -277,22 +264,24 @@ to unset the `editor.defaultFormatter` in `settings.json`:
 }
 ```
 
-### Enabling the Rust-based language server
+### Using the Rust-based language server
 
-To use the new Rust-based language server (`ruff server`), you'll need to enable the "Native Server"
-setting in the extension settings:
+The Ruff extension will automatically use the Rust-based language server (`ruff server`) if the `ruff`
+executable is at least version `0.5.3` and the `nativeServer` setting is set to `auto` (default). You can explicitly
```
MichaReiser, line 270: I think the actual implementation also considers whether the user uses any ruff-lsp only settings. Should we mention this here as well?

MichaReiser (approved):
LGTM.  Let's wait for @charliermarsh I think he wants to do the final approval.

Should we add a sentence saying that whether ruff-lsp or `ruff server` is used, can be seen in the toolbar?

dhruvmanila:
> Should we add a sentence saying that whether ruff-lsp or `ruff server` is used, can be seen in the toolbar?

Good idea. I'll add a "Troubleshooting" section because I don't think it's present in the toolbar by default.

README.md
```diff
@@ -72,40 +72,27 @@ This requires Ruff version `v0.1.3` or later.
...
-| `lint.ignore`             | `[]`          | Rules to disable by default. See [the documentation](https://docs.astral.sh/ruff/settings/#lint_ignore).                                                                    |
-| `lint.extendIgnore`       | `[]`          | Rules to disable in addition to those in `lint.ignore`.                                                                                                                     |
-| `format.preview`          | `false`       | Whether to enable Ruff's preview mode when formatting.                                                                                                                      |
+Refer to the [Ruff Language Server documentation](https://docs.astral.sh/ruff/server/settings/) for a full list of
+settings available in the extension.
+
+The extension provides the following settings in addition to those available in the Ruff Language Server that can
+be used to configure the extension. The detailed documentation for these settings can be found in the settings
+editor in Visual Studio Code.
```
charliermarsh, line 80: "Detailed documentation for..." (remove "The")

README.md
```diff
@@ -277,22 +264,26 @@ to unset the `editor.defaultFormatter` in `settings.json`:
...
 ```
 
-### Enabling the Rust-based language server
+### Using the Rust-based language server
 
-To use the new Rust-based language server (`ruff server`), you'll need to enable the "Native Server"
-setting in the extension settings:
+The Ruff extension will automatically use the Rust-based language server (`ruff server`) if the `ruff`
+executable is at least version `0.5.3`, the `nativeServer` setting is set to `auto` (default), and
+none of the settings that are specific to the Python-based language server are set. These settings
```
charliermarsh, line 271: "are enabled" over "are set"?

README.md
```diff
@@ -277,22 +264,26 @@ to unset the `editor.defaultFormatter` in `settings.json`:
...
+include the ones marked with `[1]` in the [Settings](#settings) section. You can explicitly
```
charliermarsh, line 272: "include those" instead of "include the ones"

charliermarsh (approved):

PR #535, opened 2024-07-18
Avoid passing `--preview` flag for stable `ruff server`

dhruvmanila:
## Summary

This PR updates the logic to avoid passing the `--preview` flag to the `ruff server` command when the Ruff version contains the stabilized server. 

## Test Plan

I build the latest `main` from https://github.com/astral-sh/ruff which now removes the requirement of `--preview` flag and updated the `NATIVE_SERVER_STABLE_VERSION` flag in the extension to use `0.5.2` instead.

I checked the log messages to see when it's passing the `--preview` flag.

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -61,6 +61,7 @@ export const FIND_RUFF_BINARY_SCRIPT_PATH = path.join(
 export const RUFF_SERVER_SUBCOMMAND = "server";
 
 /**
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):

dhruvmanila:
I'm going to do the same at https://github.com/neovim/nvim-lspconfig once the new version is released.

PR #539, opened 2024-07-18
Use `spawn` rather than `exec` to support paths with spaces

charliermarsh:
## Summary

If the interpreter path contains a space, then the find script fails. We need to escape the arguments that we pass to `exec`. But, it seems easier to just use `spawn`, which doesn't require escaping.

Closes https://github.com/astral-sh/ruff/issues/12394.

## Test Plan

Created a virtual environment `foo bar` in a repo; verified that the script failed before but succeeds after this change.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -38,7 +38,7 @@ import {
 import { updateServerKind, updateStatus } from "./status";
 import { getProjectRoot } from "./utilities";
 import { isVirtualWorkspace } from "./vscodeapi";
-import { exec } from "child_process";
+import { spawn } from "child_process";
 import which = require("which");
 
 export type IInitializationOptions = {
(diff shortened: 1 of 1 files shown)


src/common/server.ts
```diff
@@ -49,23 +49,40 @@ export type IInitializationOptions = {
 /**
  * Function to execute a command and return the stdout.
  */
-function executeCommand(command: string): Promise<string> {
+function executeCommand(cmd: string, args: string[] = []): Promise<string> {
   return new Promise((resolve, reject) => {
-    exec(command, (error, stdout, _) => {
-      if (error) {
-        reject(error);
+    const child = spawn(cmd, args);
```
MichaReiser, line 54: I think execFile is what we want here https://nodejs.org/api/child_process.html

charliermarsh: Oh yeah, that looks better.

charliermarsh: I can change.

flying-sheep:
I was about to file an issue for that, thanks!

Although I’m mystified why “slap a string together and execute it as a shell script” is still a thing people occasionally do. Sorry for the harshness in the preceding sentence, I appreciate all the hard work!

dhruvmanila:
> Although I’m mystified why “slap a string together and execute it as a shell script” is still a thing people occasionally do.

Our use-case is to find out the `ruff` binary which might be installed in the current virtual environment. We do this by asking the Python interpreter which requires [this script](https://github.com/astral-sh/ruff-vscode/blob/main/bundled/tool/find_ruff_binary_path.py). Another use-case is to find out the `ruff` version which is required for certain functionality like automatically switching between `ruff server` and `ruff-lsp`.

flying-sheep:
What I’m saying is that there’s a reasonable API to call a binary with optional arguments `(prog: string|path, args: (string|path)[])` and there is an error-prone way that often leads to bugs and security holes `(commandline: string)`

This PR switched from the latter to the former, for which I’m grateful, but I’m puzzled why anyone would ever default to the latter way.

PR #544, opened 2024-07-19
Use `execFile` rather than `spawn`

charliermarsh:
## Summary

Recommended in #539.


MichaReiser (approved):

src/common/server.ts
```diff
@@ -49,32 +49,15 @@ export type IInitializationOptions = {
 /**
  * Function to execute a command and return the stdout.
  */
-function executeCommand(cmd: string, args: string[] = []): Promise<string> {
+function executeCommand(file: string, args: string[] = []): Promise<string> {
```
MichaReiser, line 52: Maybe rename to executeFile

charliermarsh: Good idea.

Issue #551, opened 2024-07-23
VS Code setting ruff.interpreter does no longer accept relative paths


ni-co-la:
With the latest version of the VS Code extension (v2024.34.0 and v2024.32.0), relative paths for "ruff.interpreter" are no longer accepted (in .vscode/settings.json). 

In our repository, this settings file is shared between for all developers and it points to the venv Python installation inside the repository.
For example: 
```
    "ruff.interpreter": [
        ".venv/bin/python"
    ],
```
This was working just fine in previous versions and is still working fine with other extension like pylint and flake8.

Output:
```
2024-07-23 10:35:41.066 [error] Python version undefined.undefined is not supported.
2024-07-23 10:35:41.066 [error] Selected python path: undefined
2024-07-23 10:35:41.066 [error] Supported versions are 3.7 and above.
```

Setup:
- Windows and Linux
- Python: 3.12
- Ruff: 0.5.4
- VS Code extension: v2024.34.0
- VS Code 1.91.1

charliermarsh:
Thanks, will fix!

charliermarsh:
I am surprised that `"${workspaceFolder}/.venv/bin/python"` doesn't work.

dhruvmanila:
Can you tell me which language server are you using? Are you using the native server or `ruff-lsp` i.e., have you set the `ruff.nativeServer` setting?

I'm asking because there shouldn't be any change if you're using `ruff-lsp`. The meaning of the setting is only changed for the native server where it's only used to find out the `ruff` binary.

If it's the native server then I think it might be related to https://github.com/astral-sh/ruff-vscode/commit/e1777e138865453199626bb51cc6b2842e16360c or https://github.com/astral-sh/ruff-vscode/commit/564a37a0128cf8e55b79e260bb12f74d470eeb0d because earlier we would just pass the interpreter path as `ServerOptions` to create the `LanguageClient` in VS Code, as is still being done for `ruff-lsp`:

https://github.com/astral-sh/ruff-vscode/blob/7896831184d161ca4ab93c93b9db57eaf1609a6c/src/common/server.ts#L210-L263

charliermarsh:
We have to expand `settings.interpreter[0]`.

ni-co-la:
> I am surprised that "${workspaceFolder}/.venv/bin/python" doesn't work.

Me too. I expected that this should not make a difference for ruff, if I enter the absolute path or if I use the ${workspaceFolder}.

[...]

PR #553, opened 2024-07-23
Use `ruff.interpreter` from workspace settings

dhruvmanila:
## Summary

This PR updates the scope of `ruff.interpreter` setting from `window` to `resource` in order to resolve the variables such as `${workspaceFolder}` if present.

In #551, what happens is that the Python extension fails to resolve the environment corresponding to that interpreter which causes the hang. The bug present in #551 was there for a very long time but it got visible because of https://github.com/astral-sh/ruff-vscode/commit/e665ec7be83b72f7586b30387cd87c8faf62b9e1. Before that commit, the extension would just move ahead and use the interpreter from the Python extension but now we explicitly stop moving ahead if it fails to resolve the environment corresponding to the interpreter.

For reference, the `vscode-black-formatter` extension also updated the scope of `ruff.interpreter` to be `resource` in https://github.com/microsoft/vscode-black-formatter/commit/5f2fc61b64ee5108fe5c25d2e7d55704051256a6.

fixes: #551

### Relative paths are unresolved

```json
{
  "ruff.importStrategy": "fromEnvironment",
  "ruff.interpreter": [".venv/bin/python"]
}
```

Logs:

```
2024-07-24 08:50:37.182 [info] Using interpreter: .venv/bin/python
2024-07-24 08:50:39.240 [error] Unable to find any Python environment for the interpreter path: .venv/bin/python
```

Preview for the status bar:

Screenshot 2024-07-24 at 08 51 15

### Using VS Code specific variables

```json
{
  "ruff.importStrategy": "fromEnvironment",
  "ruff.interpreter": ["${workspaceFolder}/.venv/bin/python"]
}
```

Logs:
```
2024-07-24 08:51:43.272 [info] Using interpreter: /Users/dhruv/playground/ruff/.venv/bin/python
2024-07-24 08:51:43.306 [info] Using the Ruff binary: /Users/dhruv/playground/ruff/.venv/bin/ruff
2024-07-24 08:51:43.310 [info] Resolved 'ruff.nativeServer: auto' to use the native server
2024-07-24 08:51:43.313 [info] Found Ruff 0.5.4 at /Users/dhruv/playground/ruff/.venv/bin/ruff
2024-07-24 08:51:43.313 [info] Server run command: /Users/dhruv/playground/ruff/.venv/bin/ruff server
2024-07-24 08:51:43.313 [info] Server: Start requested.
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,6 @@ import { ExecuteCommandRequest, LanguageClient } from "vscode-languageclient/nod
 import { registerLogger, traceError, traceLog, traceVerbose } from "./common/log/logging";
 import {
   checkVersion,
-  getInterpreterDetails,
   initializePython,
   onDidChangePythonInterpreter,
   resolveInterpreter,
@@ -12,6 +11,7 @@ import { restartServer } from "./common/server";
 import {
   checkIfConfigurationChanged,
   getInterpreterFromSetting,
+  getWorkspaceSettings,
   ISettings,
 } from "./common/settings";
 import { loadServerDefaults } from "./common/setup";
@@ -23,6 +23,7 @@ import {
   onDidGrantWorkspaceTrust,
   registerCommand,
 } from "./common/vscodeapi";
+import { getProjectRoot } from "./common/utilities";
 
 const issueTracker = "https://github.com/astral-sh/ruff/issues";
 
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -263,10 +266,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         traceLog(`Python extension loading`);
         await initializePython(context.subscriptions);
         traceLog(`Python extension loaded`);
+        return; // The `onDidChangePythonInterpreter` event will trigger the server start.
       }
-    } else {
-      await runServer();
     }
+    await runServer();
   });
```
dhruvmanila, line 273: This was a bug that would hang the extension when `ruff.interpreter` is provided and the workspace is trusted.

charliermarsh (approved):

charliermarsh:
> Before that commit, the extension would just move ahead and use the interpreter from the Python extension but now we explicitly stop moving ahead if it fails to resolve the environment corresponding to the interpreter.

Do we need to fall back to the interpreter from the Python extension?

dhruvmanila:
> Do we need to fall back to the interpreter from the Python extension?

We do fall back to that, it's just moved to `settings.ts` when we get the workspace settings: https://github.com/astral-sh/ruff-vscode/pull/553/files#diff-3a5b47a3d2e56af22ee3eeb11a7ac02b69ac7b6bf52f9b48a870acd00da9f39a

PR #557, opened 2024-07-24
Avoid awaiting VS Code popup notifications

dhruvmanila:
## Summary

This PR fixes a bug to avoid awaiting the VS Code popup notifications.

### How to reproduce this?

For posterity, this is how to reproduce this bug:
1. Keep the extension setting to only contain:
```json
{
	"ruff.nativeServer": true
}
```
2. Add the following additional setting which is incompatible with the native server:
```json
{
	"ruff.nativeServer": true,
	"ruff.format.args": ["--line-length=10"]
}
```
3. Saving the new settings will trigger a restart and a notification will popup which warns you about this incompatible setting (**Do NOT dismiss the notification window**)
4. Remove the `ruff.format.args` and save the settings which should then trigger a restart again

But, the restart triggered by (3) never completed because the notification was awaited and the user never dismissed it. This is where the extension hangs.

### Solution

I've updated all `vscode.window.show(Error|Warning)Message` calls to not be awaited and instead use the callback mechanism if there's a need to respond to user's selection.

## Test Plan

Follow the same steps as above and make sure that the restart happens.

https://github.com/user-attachments/assets/28781673-9472-4ad7-9da8-d15302877307

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -115,7 +115,7 @@ async function findRuffBinaryPath(
     const stdout = await executeFile(settings.interpreter[0], [FIND_RUFF_BINARY_SCRIPT_PATH]);
     ruffBinaryPath = stdout.trim();
   } catch (err) {
-    await vscode.window
+    vscode.window
       .showErrorMessage(
         "Unexpected error while trying to find the Ruff binary. See the logs for more details.",
         "Show Logs",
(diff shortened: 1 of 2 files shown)


src/common/server.ts
```diff
@@ -246,15 +246,16 @@ async function createLegacyServer(
...
-  const selection = await vscode.window.showWarningMessage(message, "Show Logs");
-  if (selection) {
-    outputChannel.show();
-  }
+function showWarningMessageWithLogs(message: string, outputChannel: LogOutputChannel) {
+  vscode.window.showWarningMessage(message, "Show Logs").then((selection) => {
+    if (selection) {
+      outputChannel.show();
+    }
+  });
```
MichaReiser, line 254: Nit: Maybe add an error handler that logs a message if showing the message box fails for whatever reason
```suggestion
  vscode.window.showWarningMessage(message, "Show Logs").then((selection) => {
    if (selection) {
      outputChannel.show();
    }
  }).catch(ex => logError("Failed to show warning message: {ex}"));
```

MichaReiser (approved):

PR #558, opened 2024-07-24
Use `try ... finally` to always reset debounce state

dhruvmanila:
## Summary

This PR is a refactor to use `try .. finally` in the logic which starts the server to always reset the debounce state variable (`restartInProgress`). This is to ensure that it's always reset even in case of a failure.

## Test Plan

Refer to the test plan in https://github.com/astral-sh/ruff-vscode/pull/559.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -88,58 +88,57 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 
     restartInProgress = true;
 
-    const projectRoot = await getProjectRoot();
-    const workspaceSettings = await getWorkspaceSettings(serverId, projectRoot);
+    try {
+      const projectRoot = await getProjectRoot();
+      const workspaceSettings = await getWorkspaceSettings(serverId, projectRoot);
 
-    if (vscode.workspace.isTrusted) {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -88,58 +88,58 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+        outputChannel,
+        lsClient,
+      );
 
-    restartInProgress = false;
-    if (restartQueued) {
-      restartQueued = false;
-      await runServer();
+      if (restartQueued) {
+        restartQueued = false;
```
MichaReiser, line 137: I think we might still need to set `restartInProgress` to `false` before calling into `runServer` or the function will exit early. The alternative is to split the function even more: One with the `restartInProgress`  check (e.g. `debouncedRestartServer` and one without.

dhruvmanila: I'm confused a bit. Can you expand in which case it could be `false`? I'm unable to follow here.

dhruvmanila: The queued restart block has been moved into the `finally` block where it should've been which resolves the race condition I was hitting (shared on Discord).

PR #559, opened 2024-07-24
Separate server start / stop logic

dhruvmanila:
## Summary

Follow-up to #558

This PR separates the server start and stop logic.

This is required because the extension could skip starting the server in the following cases:
1. No Python interpreter is selected
2. Python extension is unable to resolve the environment for the given Python interpreter
3. Python version is incompatible

So, if the extension was already running and the settings were updated to hit any one of the above cases, the server would still be running (not ideal).

## Test Plan

Here, the diagnostics should disappear because the server stopped first before checking for the interpreter.

For (1),

I'd need to uninstall all Python versions for me to reproduce this so I'm skipping this scenario assuming that (2) and (3) is sufficient.

For (2),

https://github.com/user-attachments/assets/124ab434-c082-4e67-a606-24040d8f04a8

For (3),

https://github.com/user-attachments/assets/2300ee88-01be-45be-9c03-ae2ff4aeda78

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,7 +7,7 @@ import {
   onDidChangePythonInterpreter,
   resolveInterpreter,
 } from "./common/python";
-import { restartServer } from "./common/server";
+import { startServer, stopServer } from "./common/server";
 import {
   checkIfConfigurationChanged,
   getInterpreterFromSetting,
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -269,6 +272,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 
 export async function deactivate(): Promise<void> {
   if (lsClient) {
-    await lsClient.stop();
+    await stopServer(lsClient);
   }
 }
```
dhruvmanila, line 277: Also, this seems like a bugfix because otherwise the listeners wouldn't be disposed.

Issue #583, opened 2024-08-15
Extension breaks on Windows if Ruff binary file path contains non-ASCII characters


floxay:
There are actually two issues on Windows:
1) https://github.com/astral-sh/ruff-vscode/blob/main/bundled/tool/find_ruff_binary_path.py fails if it encounters a character not supported by Windows-1252 (cp1252), let's say if I have a directory called `löカひ漢` it will fail with the trace below;
	<details>
	<summary>Python Trace:</summary>

	```py
	2024-08-15 13:59:10.579 [info] Using interpreter: c:\löカひ漢\.venv\Scripts\python.exe
	2024-08-15 13:59:10.735 [error] Error while trying to find the Ruff binary: Error: Traceback (most recent call last):
	  File "c:\Users\user\.vscode\extensions\charliermarsh.ruff-2024.40.0-win32-x64\bundled\tool\find_ruff_binary_path.py", line 35, in <module>
	    print(os.fsdecode(str(ruff_binary_path)), flush=True)
	  File "C:\Program Files\Python312\Lib\encodings\cp1252.py", line 19, in encode
	    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
	           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	UnicodeEncodeError: 'charmap' codec can't encode characters in position 5-7: character maps to <undefined>
	```
	</details>
2) Non-ASCII characters which are part of Windows-1252 prevent the extension from using the Ruff binary, the extension will be stuck in a "loading" phase;
	```py
	2024-08-15 14:10:49.461 [info] Using interpreter: c:\lö\.venv\Scripts\python.exe
	2024-08-15 14:10:49.585 [info] Using the Ruff binary: c:\l�\.venv\Scripts\ruff.exe
	```
	Relevant line and function: https://github.com/astral-sh/ruff-vscode/blob/d84e14b6326b28997920bbc23f5a89b983f0effc/src/common/server.ts#L129

MichaReiser:
Thanks. I can reproduce this. The bug is in `find_ruff_binary_path.py` and only happens if there's a ruff binary in the venv. The error doesn't arise if the venv does not contain a ruff binary. 

To unblock you, you can specify `ruff.path`  and set it to `C:\löカひ漢\.venv\Scripts\ruff.exe`.

MichaReiser:
[...]

PR #584, opened 2024-08-15
Fix ruff discovery for paths containing symbols that can't be represented by the local encoding

MichaReiser:
## Summary

Fixes https://github.com/astral-sh/ruff-vscode/issues/583

See inline comment for an explanation

## Test Plan

* Created a local directory named `löカひ漢`
* Added a venv
* installed ruff

Before fix: The python script crashed
After: It finds the right ruff binary


bundled/tool/find_ruff_binary_path.py
```diff
@@ -32,4 +39,4 @@ def find_ruff_binary_path() -> Optional[Path]:
 if __name__ == "__main__":
     ruff_binary_path = find_ruff_binary_path()
     if ruff_binary_path:
-        print(os.fsdecode(str(ruff_binary_path)), flush=True)
+        print(os.fsdecode(ruff_binary_path), flush=True)
```
MichaReiser, line 42: I think the call to `str` before calling `os.fsdecode` is just wrong? But how would I know, I don't know nothing of Python

dhruvmanila: Yeah, I think that's correct. `print(<pathlib.Path>)` would use the `pathlib.Path.__str__` method, so the `str` call here is redundant.

dhruvmanila (approved):
Thanks!

bundled/tool/find_ruff_binary_path.py
```diff
@@ -30,6 +30,13 @@ def find_ruff_binary_path() -> Optional[Path]:
 
 
 if __name__ == "__main__":
+    # Python defaults to the system's local encoding for stdout on Windows.
+    # source: https://docs.python.org/3/library/sys.html#sys.stdout
+    #
+    # But not all paths are representable by the local encoding.
+    # The node process calling this script defaults to UTF8, so let's do the same here.
+    sys.stdout.reconfigure(encoding="utf-8")  # type: ignore We never reconfigure stdout, thus it is guaranteed to not be Any
```
dhruvmanila, line 38:
```suggestion
    sys.stdout.reconfigure(encoding="utf-8")  # type: ignore [union-attr] # We never reconfigure stdout, thus it is guaranteed to not be Any
```

PR #600, opened 2024-09-05
Always include "Show Logs" button for window messages

dhruvmanila:
## Summary

This PR includes a "Show Logs" button for all window messages from Ruff. This might be useful to reduce the friction for the users in figuring out where the logs are. This is something I saw when looking at https://github.com/koxudaxi/ruff-pycharm-plugin/issues/463 description.

## Test Plan

Screenshot 2024-09-06 at 10 15 55

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -2,7 +2,7 @@ import * as fsapi from "fs-extra";
 import * as vscode from "vscode";
 import { platform } from "os";
 import { Disposable, l10n, LanguageStatusSeverity, LogOutputChannel } from "vscode";
-import { State } from "vscode-languageclient";
+import { State, ShowMessageNotification, MessageType } from "vscode-languageclient";
 import {
   LanguageClient,
   LanguageClientOptions,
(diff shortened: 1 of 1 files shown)


dhruvmanila:
This isn't the complete picture because there's still https://github.com/astral-sh/ruff/issues/12523#issuecomment-2333238119 that needs to be solved.

src/common/server.ts
```diff
@@ -454,14 +462,20 @@ export async function startServer(
           break;
       }
     }),
+    newLSClient.onNotification(ShowMessageNotification.type, (params) => {
```
MichaReiser, line 465: For my understanding. This doesn't apply to `$logTrace`? This is only run when the LSP sends an explicit show message request?

I'm asking because I think that showing a dialog for info and warning traces would be a bit overly aggressive.

MichaReiser (approved):

src/common/server.ts
dhruvmanila: Yeah, this doesn't apply to `$/logTrace`. This only applies to the https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage request which is what the editor uses to display notifications from the server.

Issue #605, opened 2024-09-07
VSCode status bar display is weird in the latest (2024.44.0)


ryokaji:
## What happens
When `2024.44.0` is installed, the Status bar display in the default (v0.6.4) looks like hunging up as follows.
![image](https://github.com/user-attachments/assets/3c5146dd-ea99-4a1f-a4d2-780587691b3a)

This does not occur if Ruff v0.5.7 is explicitly specified in the settings.
![image](https://github.com/user-attachments/assets/aba366b3-feef-4a47-93b5-e9a536abb854)

*note: This issue is not reporting a bug, the extension looks working fine.

## System Info
VSCode version: 1.91.1
Python version: 3.11.2
Host OS:
* Microsoft Windows 10 Home 22H2 19045.4780
* macOS Sonoma 14.5

## Project structure
The following stuff is minimum configuration to reproduce the issue.
### Settings
```json5
{
    // Configuration to explicitly specify v0.5.7
    // Assuming that a venv has been created in ~/poc.
    "ruff.path": [
        "/home/node/poc/bin/ruff"
    ]
}
```

### Others

files:
```
.
├── hello.py
└── .vscode
    └── settings.json
```

hello.py
```python
print("hello")

```

MichaReiser:
Thanks. I can reproduce this by just opening a python project.

PR #608, opened 2024-09-08
Set status to 'not-busy' when server startup is complete

MichaReiser:
## Summary

Ensures that the extension calls `updateStatus` with `busy=false` when the server startup is complete. 

This is now necessary after https://github.com/astral-sh/ruff-vscode/pull/600 correctly added an `updateStatus` call with `busy=true` before the server starts.

Fixes https://github.com/astral-sh/ruff-vscode/issues/605

## Test Plan

I ran the extension and verified that the busy indicator no longer spins forever.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -441,6 +441,7 @@ export async function startServer(
   traceInfo(`Server: Start requested.`);
   try {
     await newLSClient.start();
+    updateStatus(undefined, LanguageStatusSeverity.Information, false);
   } catch (ex) {
     updateStatus(l10n.t("Server failed to start."), LanguageStatusSeverity.Error);
     traceError(`Server: Start failed: ${ex}`);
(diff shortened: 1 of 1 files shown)


dhruvmanila:
I'm confused, where did #600 add the `updateStatus` call? I think that PR only changed the ordering such that the disposables are added only after a successful server start.

MichaReiser:
Hmm good call. I have to look into this again what the root cause is for it breaking

dhruvmanila:
The bug was that the `onDidChangeState` handler was being added after the server started so when the client updated the status to running, the handler wasn't present which would've changed the status. I've pushed a commit that changed the order back and added a helper function to dispose and remove them from the array and is being used when (a) the server failed to start or (b) when stopping the server.

dhruvmanila (approved):

PR #614, opened 2024-09-10
Separate publish steps for Code Marketplace and OpenVSX

dhruvmanila:
## Summary

Publishing the extension to the OpenVSX registry fails a lot of times and there's no way to re-run just the publish step for the OpenVSX in GitHub Actions, we can only re-run the entire workflow. This PR separates the publish step into two where (1) would publish to Code Marketplace and (2) would publish to OpenVSX registry. This means that if we fail to publish to one of the registry, we could still re-run it.

### Limitation

Sometimes it does happen that only some of the artifacts are published and not all of them in which case re-running won't solve this issue. I don't really have a good solution here.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -166,8 +166,8 @@ jobs:
           path: ./dist
 
   # Phase 3: Publish the built extension to the Marketplace.
-  publish:
-    name: "Publish"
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Thanks @dhruvmanila. This is great

PR #631, opened 2024-10-18
Pin `ovsx` as a dev dependency

dhruvmanila:
## Summary

The recent version of `ovsx` had a breaking change which bumps the minimum Node version required to be 20. This broke our release workflow: https://github.com/astral-sh/ruff-vscode/actions/runs/11392480125/job/31698800337.

Similar to how we've pinned the `@vscode/vsce` which is a VS Code extension manager and is used to publish the artifacts to the VS Code Marketplace, we'll pin the `ovsx` as a dev dependency. Both had a breaking change but as the latter wasn't pinned, it used the latest version in CI.

Currently, the `ovsx` is pinned to `0.9.4` as `0.9.5` had the breaking change: https://github.com/eclipse/openvsx/blob/master/cli/CHANGELOG.md#v095-sep-2024

## Test Plan

`npx ovsx --version` should give `0.9.4`

<details><summary>Details</summary>
<p>

```console
❯ just clean
rm -rf out
rm -rf node_modules
rm -rf .vscode-test
rm -rf bundled/libs

❯ npx ovsx --version
Need to install the following packages:
ovsx@0.10.0
Ok to proceed? (y) y

0.10.0

❯ npx ovsx --version
0.10.0

❯ npm ci            
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported

added 531 packages, and audited 532 packages in 2s

116 packages are looking for funding
  run `npm fund` for details

5 vulnerabilities (3 moderate, 2 high)

To address all issues, run:
  npm audit fix

Run `npm audit` for details.

❯ npx ovsx --version
0.9.4
```

</p>
</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
         "eslint": "^8.30.0",
         "eslint-plugin-prettier": "^4.2.1",
         "glob": "^10.2.2",
+        "ovsx": "^0.9.4",
         "prettier": "^2.8.8",
         "ts-loader": "^9.4.2",
         "typescript": "^5.1.3",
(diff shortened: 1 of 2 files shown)


MichaReiser:
It's a bit unfortunate that the result is that everyone has to install `ovsx` which is only used in CI. an alternative is to use `npx -y ovsx@0.9` but it comes with the downside that renovate won't upgrade it. 

Either way. Should we upgrade our CI job to use something newer than node 20?

dhruvmanila:
> It's a bit unfortunate that the result is that everyone has to install `ovsx` which is only used in CI. an alternative is to use `npx -y ovsx@0.9` but it comes with the downside that renovate won't upgrade it.

Yeah, I initially thought of specifying the version directly in the workflow file but this was the reason I didn't do so.

> Either way. Should we upgrade our CI job to use something newer than node 20?

Yes, I can do that in a follow-up PR.

MichaReiser (approved):

PR #633, opened 2024-10-18
Update Node version in release workflow

dhruvmanila:
## Summary

TODO: Need to test this

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -133,9 +133,9 @@ jobs:
 
       # Install Node.
       - name: Install Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 22
 
       # Install Node dependencies.
       - run: npm ci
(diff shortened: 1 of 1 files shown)


MichaReiser:
We could consider testing this together with upgrading the release dependencies

dhruvmanila:
> We could consider testing this together with upgrading the release dependencies

Yeah, makes sense.

MichaReiser:
Note. I think the marketplace have a `--dry-run` flag or similar that we could use to test the change without publishing.

dhruvmanila:
> Note. I think the marketplace have a `--dry-run` flag or similar that we could use to test the change without publishing.

Hmm, I don't see a `--dry-run` or related flag in the `vsce` / `ovsx` tools that we use to publish.

MichaReiser:
I guess you could try --skip-duplicate but feels risky. I'm also up to just merge and revert if we run into problems during the next release.

dhruvmanila:
> I guess you could try --skip-duplicate but feels risky. I'm also up to just merge and revert if we run into problems during the next release.

Yeah, that does seem risky. Another option would be to create a dummy account and do the release there but I think I'd just prefer to merge it and make a mental note to verify when doing the release.

MichaReiser (approved):

PR #634, opened 2024-10-18
CONTRIBUTING: `uv` is now used to generate our `requirements.txt` files

AlexWaygood:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,8 +6,9 @@ This extension is based on the [Template for VS Code Python tools extensions](ht
 
 ### Getting Started
 
+- Install [`uv`](https://github.com/astral-sh/uv)
 - Install [`just`](https://github.com/casey/just), or see the `justfile` for corresponding commands.
-- Create and activate a virtual environment (e.g., `python -m venv .venv && source .venv/bin/activate`).
+- Create and activate a virtual environment (e.g., `uv venv && source .venv/bin/activate`).
 - Install development dependencies (`just install`).
 - To automatically format the codebase, run: `just fmt`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,10 +35,9 @@ and Python log messages in the debug console under "Python Server".
 - Update the `ruff` version in the README.md
   - in the Base URLs
   - in "The extension ships with `ruff==...`"
-- Make sure you use Python 3.7 installed and as your default Python.
-- Run `python -m venv .venv` to create a venv and activate it.
-- Run `python -m pip install pip-tools` to install `pip-tools`.
-- Run `rm requirements.txt requirements-dev.txt` and then `just lock` to update `ruff` and `ruff-lsp`.
+- Make sure you have Python 3.7 installed
+- Run `uv venv --python 3.7 && source .venv/bin/activate` to create a Python 3.7 venv and activate it.
```
AlexWaygood, line 39: Unfortunately `uv` will not install Python 3.7 for you if you don't have it already installed; 3.7 is too old for `uv`

dhruvmanila (approved):
Thanks!

CONTRIBUTING.md
dhruvmanila: Ah yeah, I remember facing this

dhruvmanila: I think Python is only required to bundle `ruff-lsp` so it should be fine

AlexWaygood: Python 3.7 is also still installable by other means, so it's "okay", just makes it a bit more annoying to work with if you don't already have Python 3.7 installed locally. Luckily I do 😄

PR #647, opened 2024-11-26
Add a script for automating releases

AlexWaygood:
## Summary

This adds a script to automate the changes to various different files that need to be made in order to cut a new `ruff-vscode` release. This is one of the many things you need to do as part of the Ruff minor release process, which I'd like to make less time-consuming.

The script does the following things:
- Bumps the version of this package in `pyproject.toml` and `package.json`
- Bumps the `ruff` and `ruff-lsp` dependency pins this project has in `pyproject.toml` to the latest versions
- Makes any necessary edits to the README
- Adds a draft changelog entry
- Updates the package's lockfiles
- Optionally commits the changes to a new branch

## Test Plan

I checked out 51bc650b10a8662cc062667e7ff6959a8314da2f and ran `uv run --python=3.7 release.py`. The changes the script made were almost identical to https://github.com/astral-sh/ruff-vscode/commit/6ac401c6474f0acfa2d28a04aeef7ca6b4261af5 (the latest commit we had preparing a new release).

I also ran the following command to make sure the script type-checked with strict settings:

```
uvx --with=types-requests --with=tomli --with=tomlkit --with=packaging --with=rich-argparse mypy release.py --strict --warn-unreachable --enable-error-code=possibly-undefined --enable-error-code=redundant-expr --enable-error-code=truthy-bool
```


CONTRIBUTING.md
```diff
@@ -30,15 +30,12 @@ and Python log messages in the debug console under "Python Server".
...
-  - in the Base URLs
-  - in "The extension ships with `ruff==...`"
-- Make sure you have Python 3.7 installed
-- Run `uv venv --python 3.7 && source .venv/bin/activate` to create a Python 3.7 venv and activate it.
-- Run `rm requirements.txt requirements-dev.txt && just lock` to update `ruff` and `ruff-lsp`.
-- Update the Changelog
+- Make sure you have Python 3.7 installed and locatable by uv.
+  (If you're using pyenv, you may need to run `pyenv local 3.7`.)
+- Run `uv run --python=3.7 scripts/release.py`.
+  (Run `uv run --python=3.7 scripts/release.py --help` for information on what this script does,
```
MichaReiser, line 36: Could you add this as a `just` script. E.g. `just release`?

scripts/release.py
```diff
@@ -0,0 +1,244 @@
...
+def get_ruff_versions() -> RuffVersions:
+    """
+    Obtain metadata about the project; figure out what the new metadata should be.
+    """
+    with PYPROJECT_TOML_PATH.open("rb") as pyproject_file:
+        pyproject_toml = tomli.load(pyproject_file)
+
+    existing_vscode_version = pyproject_toml["project"]["version"]
+
+    major, minor, micro = Version(existing_vscode_version).release
```
MichaReiser, line 80:
```suggestion
    major, minor, patch = Version(existing_vscode_version).release
```

scripts/release.py
```diff
@@ -0,0 +1,244 @@
...
+    new_vscode_version = Version(f"{major}.{minor + 2}.{micro}")
```
MichaReiser, line 81: Hmm, how would we do a patch release or release a preview version (which we've done a few times in the past)?

scripts/release.py
```diff
@@ -0,0 +1,244 @@
...
+            f"version to `{versions.latest_ruff_lsp}`."
+        )
+    elif versions.latest_ruff != versions.existing_ruff_pin:
+        changelog_entry_middle = f"This release upgrades the bundled Ruff version to `v{versions.latest_ruff}`."
+    elif versions.latest_ruff_lsp != versions.existing_ruff_lsp_pin:
+        changelog_entry_middle = f"This release upgrades the bundled `ruff-lsp` version to `v{versions.latest_ruff_lsp}`."
+    else:
+        changelog_entry_middle = ""
+
+    changelog_entry = textwrap.dedent(f"""\
```
MichaReiser, line 172: Why do you need the `\`. 

Isn't it nice how Ruff wraps the multiline string? :)

MichaReiser (approved):
I haven't reviewed the script in detail. I'm sure it works. 

I suggest that we add CLI options for all *new versions* to support future patch releases, preview releases, and any unforseen scenario

scripts/release.py
AlexWaygood: It's a very common practice when using `textwrap.dedent()` in combination with triple-quoted strings, or it's easy to end up with a dedented string where the first line has odd indentation https://grep.app/search?q=textwrap.dedent%28%22%22%22%5C&filter[lang][0]=Python

> Isn't it nice how Ruff wraps the multiline string? :)

I'm... actually not a massive fan of that formatting choice 🙈 but I don't care enough to argue about it 😆

scripts/release.py
AlexWaygood: My idea of this script was to cover the _common_ workflow rather than _all possible_ workflows. I can add a CLI flag for this if you want, but I worry that the complexity it will add to the script won't be worth the time it will save us

scripts/release.py
MichaReiser: I know it is... but it's something black doesn't support :)

scripts/release.py
MichaReiser: The problem I see is that this PR removes the documentation on how to do a release manually when not using the script. To me, that means it's now the only way of doing a release. 

That's why I think we should either revert the documentation changes or extend the script. I'm worried about reverting the documentation because it probably will get outdated with future changes to this script which then is even worse: You end up with incomplete/incorrect instructions

AlexWaygood: I see. Okay, I'll add the CLI flags.

AlexWaygood: I added more flags in https://github.com/astral-sh/ruff-vscode/pull/647/commits/37b29352724609e8f8fa1926e4043905fea2376d, and tested them using `uv run --python=3.7 scripts/release.py --new-version=2024.600.0 --new-ruff=0.6.0`. It did everything I expected it to.

Also, it didn't end up being much more code at all, so I think it _was_ worth it -- thanks for persuading me!

CONTRIBUTING.md
AlexWaygood: I've added this. I haven't documented it as "the recommended way" to create a release, though, as it doesn't look like there's a way to pass the various optional flags (`--help`, `--new-version`, `--new-ruff`, etc.) when you run it as a `just` script. justfiles don't seem to support optional arguments (only arguments with default values, which isn't _quite_ what I want here).

AlexWaygood:
Thanks for the review!

scripts/release.py
MichaReiser: Thanks for adding it and thanks for automating the steps!

PR #648, opened 2024-11-27
release script: default the major version to the current year, not the existing major version

AlexWaygood:
We use a form of CalVer, so this makes more sense


AlexWaygood:
oops, I assumed pressing that button would enable automerge rather than just insta-merging!

PR #659, opened 2025-01-03
Improve logging system using separate output channels

dhruvmanila:
## Summary

Related: https://github.com/astral-sh/ruff/pull/15232

The basic idea is that there will be two channels where the logs will go:
1. Ruff: Here, all the logs for the extension will go
2. Ruff Language Server: Here, all the server log messages will go and are captured from the stderr. It will also contain the trace communications between the client and the server when it's enabled.
3. Ruff Language Server Trace: Here, all the communication messages between the client and the server will go. **This channel is created lazily when the value of `ruff.trace.server` is changed to "verbose" or "messages" for the first time.**

(2) is controlled by the `ruff.logLevel` configuration which defaults to `info`.

(3) is controlled by the `ruff.trace.server` configuration which defaults to `off` and can be changed to `messages` and `verbose`. The difference between them is that the former will only log the method name for both the request and response while the latter will also log the request parameters and the response result.

### Red Knot

The end goal is to merge the tracing system with Red knot. Currently, Red knot's tracing system utilizes the `-v`, `-vv` and `-vvv` flags from the command-line and `RED_KNOT_LOG` environment variable for configuration. For the red knot server specifically, we will need to provide an additional configuration parameter like `ruff.server.extraEnv` which allows user to fine tune the logging using the mentioned environment variable. The `ruff.logLevel` will by default pass `RED_KNOT_LOG=<logLevel>` to allow for easy configuration.

### Why not `window/logMessage` ?

The `window/logMessage` request can be used by the server to notify the client to log a message at a certain log level. There are a few disadvantages of using this as oppose to just using stderr:
* It needs to go through the RPC which requires serializing and de-serializing the messages
* The output format depends on how client handles the message
* The trace channel will get populated with "Received notification 'window/logMessage'" lines 
* There's no 'debug' and 'trace' message type the LSP protocol although the next version is going to include debug type

Additionally, putting the message onto stderr also allows us to use a custom formatter which is what is being done in red knot and would prove to be beneficial when merging the tracing setup.

<details><summary>For posterity, here's the patch to use <code>window/logMessage</code>:</summary>
<p>

```diff
diff --git a/crates/ruff_server/src/trace.rs b/crates/ruff_server/src/trace.rs
index 7b367c780..ceedeff21 100644
--- a/crates/ruff_server/src/trace.rs
+++ b/crates/ruff_server/src/trace.rs
@@ -15,6 +15,9 @@
 //! Tracing will write to `stderr` by default, which should appear in the logs for most LSP clients.
 //! A `logFile` path can also be specified in the settings, and output will be directed there instead.
 use core::str;
+use lsp_server::{Message, Notification};
+use lsp_types::notification::{LogMessage, Notification as _};
+use lsp_types::{LogMessageParams, MessageType};
 use serde::Deserialize;
 use std::{
     path::PathBuf,
@@ -22,6 +25,7 @@ use std::{
     sync::{Arc, OnceLock},
 };
 use tracing::level_filters::LevelFilter;
+use tracing_subscriber::fmt::MakeWriter;
 use tracing_subscriber::{
     fmt::{time::Uptime, writer::BoxMakeWriter},
     layer::SubscriberExt,
@@ -73,7 +77,7 @@ pub(crate) fn init_tracing(
 
     let logger = match log_file {
         Some(file) => BoxMakeWriter::new(Arc::new(file)),
-        None => BoxMakeWriter::new(std::io::stderr),
+        None => BoxMakeWriter::new(LogMessageMakeWriter),
     };
     let subscriber = tracing_subscriber::Registry::default().with(
         tracing_subscriber::fmt::layer()
@@ -135,3 +139,61 @@ impl<S> tracing_subscriber::layer::Filter<S> for LogLevelFilter {
         Some(LevelFilter::from_level(self.filter.trace_level()))
     }
 }
+
+struct LogMessageMakeWriter;
+
+impl<'a> MakeWriter<'a> for LogMessageMakeWriter {
+    type Writer = LogMessageWriter;
+
+    fn make_writer(&'a self) -> Self::Writer {
+        LogMessageWriter {
+            level: tracing::Level::INFO,
+        }
+    }
+
+    fn make_writer_for(&'a self, meta: &tracing::Metadata<'_>) -> Self::Writer {
+        LogMessageWriter {
+            level: *meta.level(),
+        }
+    }
+}
+
+struct LogMessageWriter {
+    level: tracing::Level,
+}
+
+impl LogMessageWriter {
+    const fn message_type(&self) -> MessageType {
+        match self.level {
+            tracing::Level::ERROR => MessageType::ERROR,
+            tracing::Level::WARN => MessageType::WARNING,
+            tracing::Level::INFO => MessageType::INFO,
+            tracing::Level::DEBUG => MessageType::LOG,
+            tracing::Level::TRACE => MessageType::LOG,
+        }
+    }
+}
+
+impl std::io::Write for LogMessageWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let message = str::from_utf8(buf)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?
+            .to_owned();
+        LOGGING_SENDER
+            .get()
+            .expect("logging sender should be initialized at this point")
+            .send(Message::Notification(Notification {
+                method: LogMessage::METHOD.to_owned(),
+                params: serde_json::to_value(LogMessageParams {
+                    typ: self.message_type(),
+                    message,
+                })?,
+            }))
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
```

</p>
</details> 

## Test Plan

Using the following **VS Code** settings:

```json
{
  "ruff.nativeServer": "on",
  "ruff.path": ["/Users/dhruv/work/astral/ruff/target/debug/ruff"],
  "ruff.trace.server": "verbose",
  "ruff.logLevel": "debug"
}
```

The following video showcases the three channels in VS Code:

https://github.com/user-attachments/assets/53396774-9dfb-4c33-9145-67eaf2395936

Using the following **Zed** settings:

```json
{
  "lsp": {
    "ruff": {
      "binary": {
        "path": "/Users/dhruv/work/astral/ruff/target/debug/ruff",
        "arguments": [
          "server"
        ]
      },
      "initialization_options": {
        "settings": {
          "logLevel": "debug"
        }
      }
    }
  }
}
```

The following video showcases that the stderr messages are captured in the "Server Logs" window and the RPC window provides the raw request and response messages:

https://github.com/user-attachments/assets/ee36eb4a-133f-46d3-abb9-366c48793f32

For **Neovim**, there's an open issue (https://github.com/neovim/neovim/issues/16807) to send stderr messages to a separate file but currently it gets logged as "ERROR" in the main LSP log file. There doesn't seem to be any way to hook into the stderr of the Neovim LSP client for us to provide a solution for Ruff users.


MichaReiser:
Thanks for working on this. I haven't reviewed the code yet because I've some UX questions. 

As a user, I think I'd find it confusing that the extension and server settings are separated. For example, Red Knot logs:

```
INFO Python version: Python 3.9
INFO Found 8 files in package `test`
error[lint:unresolved-import] /Users/micha/astral/test/distribution/distribution.py:1:8 Cannot resolve import `seaborn`
error[lint:unresolved-import] /Users/micha/astral/test/distribution/distribution.ipynb:1:8 Cannot resolve import `seaborn`
error[lint:unresolved-import] /Users/micha/astral/test/distribution/distribution.ipynb:2:8 Cannot resolve import `pandas`
error[lint:unresolved-import] /Users/micha/astral/test/other.py:2:1 Cannot resolve import `c`
```

I would find it most useful as a user when I see this in the same panel as the main extension logs. Having to switch between the two panels will make it difficult to match the log lines (what's the order of the events?)

As a developer, I think the separation of the server and tracing output might be problematic because it makes it harder to correlate events. Did the panic happen before or after this specific LSP message? I'm worried that not having this knowledge will make debugging errors harder.

Can you tell me more for the motivation for separating the two outputs?

dhruvmanila:
My main motivation behind having separate channels to so that it's easy to go through the logs which are specific to either the client (editor) or the server. Additionally, the format of the log messages in the extension channel is different because it's a log channel while the server messages uses a normal channel. The difference between a log and a normal channel in VS Code is that the former will add the timestamp and the log level while the latter will add the messages as is. This separation is useful especially for the server because the logs from both the server and the CLI output will have exactly the same format and it allows the server to be in control of the message format.

The kind of log messages between the editor and the server are also quite different considering that once the server starts, there's almost no involvement from the client side apart from notifying the user for any issues regarding the server.

> As a developer, I think the separation of the server and tracing output might be problematic because it makes it harder to correlate events. Did the panic happen before or after this specific LSP message? I'm worried that not having this knowledge will make debugging errors harder.

Yeah, I think I see your point and it makes sense to have one channel for both the server logs and the trace logs at least.

I'm interested in your opinion regarding whether to have a separate channel for client and server logs. Currently, my main worry is about the message format.

MichaReiser:
Thanks. The explanation about the different channel styles is helpful, and I can see how that can motivate having separate server and client traces. 

Looking at Red Knot, all logs with `info` are targeted at end users. To some degree, this even includes `debug` messages. It's really only `trace` messages that are only intended for developers. The same is `true` for the extension logs, which is why it would be nice if all of them could be found in a single place. I'd also say that multiple output panels are probably harder to understand than just one. E.g., as a user, I don't think I'd know which output channel I should look at if something isn't working. That's why I'd say that the fewer channels the better (as long as we find a good UX)

But then, having two channels seems to be fairly common: Both Rust analyzer and Python have separate output channels (ESLint, TypeScript have just one). Biome uses Biome and tracing

> The kind of log messages between the editor and the server are also quite different considering that once the server starts, there's almost no involvement from the client side apart from notifying the user for any issues regarding the server.

That's fair, although I don't think this matters much from a user perspective. I just want to see the logs. 

> server to be in control of the message format.

This is nice, although I think we could customize the tracing output to have a format that's optimized for the VS code output. 

Either way. I think two channels, one for the client and one for the server, seems a reasonable start.

MichaReiser:
We discussed this via chat and decided to keep the three output channels because:

* Combining the VS Code tracing output channel and the client channel has the downside that trace messages end up with two time stamps: one from the log output panel and one from the trace
* We create the trace output panel lazily. There's no point in having an empty output panel for most users. It's primarily a development feature
* We improve the server spans to create a span for each request/response that includes the request method and request ID. This allows correlating logs between the tracing and server panels.

.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
```
dhruvmanila, line 16: This is because the latest Ubuntu version doesn't provide Python 3.7

src/extension.ts
```diff
@@ -158,7 +151,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
       await runServer();
     }),
     registerCommand(`${serverId}.showLogs`, async () => {
-      outputChannel.show();
+      logger.channel.show();
     }),
```
MichaReiser, line 155: I wonder if this is now confusing. It suggests that there's only one log and a user might not be aware that there's more than one. Should we add a new `showLspLogs` command and change their lables to `Show client logs` and `Show LSP logs`?

src/common/logger.ts
```diff
@@ -0,0 +1,96 @@
+import * as util from "util";
+import * as vscode from "vscode";
+
+class ExtensionLogger {
```
MichaReiser, line 4: It would be great if we could add a short description of what this logger does and what it is used for (or to the `logger `constant below)

src/common/logger.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (this.isCI) {
+      console.log(...messages);
+    }
+    this.channel.trace(util.format(...messages));
+  }
+}
+
+export const logger = new ExtensionLogger();
+
+export class LazyOutputChannel implements vscode.OutputChannel {
```
MichaReiser, line 46: Same here: I'd find a short comment explaining what this output panel is used for and what the lazy part is useful.

MichaReiser (approved):
Thanks. This is great. 

For Neovim. I assume the proposed solution is to set `logFile`?

src/common/server.ts
```diff
@@ -220,7 +218,7 @@ async function createLegacyServer(
   settings: ISettings,
   serverId: string,
   serverName: string,
-  outputChannel: LogOutputChannel,
+  outputChannel: OutputChannel,
```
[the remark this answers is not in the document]
dhruvmanila: It's not strictly required but I don't think it would hurt to do so. I'll update it.

src/extension.ts
dhruvmanila: Good point. I think that makes sense although what about naming it "Show language server logs" / "Show server logs" instead?

MichaReiser: I don't mind "Show server logs"

dhruvmanila:
> For Neovim. I assume the proposed solution is to set `logFile`?

Yes, ideally that's the best solution for all editors that uses a single file for logs from all the language server. I'll update the Ruff documentation to add a tip for this.

PR #685, opened 2025-02-09
Improve deprecation warnings

dhruvmanila:


dhruvmanila:
I'm a bit unsure on the flow of (5) and (6), any suggestions are appreciated.

MichaReiser:
For 4: I find it a bit confusing that we suggest setting to on `nativeServer: on` when it is sufficient for them to migrate the settings. Requiring explicitly setting `nativeServer: on` does have the downside that many users never set `nativeServer: auto`, it's just the default (or did we change the default)?

For 5: I'm unsure what the info message is trying to tell me. Does it mean I have to upgrade to use the native server or is it only a hint that I'd be using an experimental version if I keep using the older version? 

6 doesn't seem ideal but it's also not terrible. Maybe we could add a note to the migration guide that using the native server requires at least Ruff 0.3.5 to give users a heads up?

dhruvmanila:
Thank you for the feedback.

> For 4: I find it a bit confusing that we suggest setting to on `nativeServer: on` when it is sufficient for them to migrate the settings. Requiring explicitly setting `nativeServer: on` does have the downside that many users never set `nativeServer: auto`, it's just the default (or did we change the default)?

Yes, "auto" is the default value. I'll remove the `nativeServer: on` suggestion.

> For 5: I'm unsure what the info message is trying to tell me. Does it mean I have to upgrade to use the native server or is it only a hint that I'd be using an experimental version if I keep using the older version?

That message already existed, but I think I can improve it:

```
Stable version of the native server requires Ruff 0.5.3, but found 0.5.2 at /Users/dhruv/.local/bin/ruff instead; using the legacy server (ruff-lsp)
```

Note that there's also this info message that'll be present so overall it says that the auto behavior found the Ruff version which does not include stable server so will use ruff-lsp instead:

```
Resolved 'ruff.nativeServer: auto' to use the legacy (ruff-lsp) server
```

> 6 doesn't seem ideal but it's also not terrible. Maybe we could add a note to the migration guide that using the native server requires at least Ruff 0.3.5 to give users a heads up?

Yeah, that makes sense.

src/common/server.ts
```diff
@@ -263,96 +267,74 @@ async function createLegacyServer(
...
-
 type RuffExecutable = {
   path: string;
   version: VersionInfo;
 };
 
+const ruffLspUrl = "https://github.com/astral-sh/ruff-lsp";
+const migrationUrl = "https://docs.astral.sh/ruff/editors/migration/";
+const gitHubDiscussionMessage =
+  "Feel free to comment on the [GitHub discussion](https://github.com/astral-sh/ruff/discussions/15991) to ask questions or share feedback.";
```
MichaReiser, line 287:
```suggestion
const RUFF_LSP_URL = "https://github.com/astral-sh/ruff-lsp";
const LSP_MIGRATION_URL = "https://docs.astral.sh/ruff/editors/migration/";
const LSP_DEPRECATION_DISCUSSION_MESSAGE =
  "Feel free to comment on the [GitHub discussion](https://github.com/astral-sh/ruff/discussions/15991) to ask questions or share feedback.";
```

src/common/server.ts
```diff
@@ -263,96 +267,74 @@ async function createLegacyServer(
...
   switch (settings.nativeServer) {
     case "on":
     case true:
-      const legacyServerSettings = getUserSetLegacyServerSettings(serverId, workspace);
+      legacyServerSettings = getUserSetLegacyServerSettings(serverId, workspace);
       if (legacyServerSettings.length > 0) {
-        legacyServerSettingsWarning(legacyServerSettings);
+        // User has explicitly set the native server to 'on' but still has legacy server settings.
+        showWarningMessage(
+          `Following settings have been deprecated in the native server: ${JSON.stringify(
```
MichaReiser, line 305:
```suggestion
          `The following settings have been deprecated in the native server: ${JSON.stringify(
```

src/common/server.ts
```diff
@@ -361,38 +343,39 @@ async function resolveNativeServerSetting(
...
+        // User has legacy server settings set.
         useNativeServer = false;
       }
 
       if (!useNativeServer) {
-        // For versions >= 0.3.5, `ruff-lsp` will show the deprecation warning.
-        if (!supportsNativeServer(ruffVersion)) {
-          legacyServerDeprecationWarning(ruffVersion);
+        let message =
+          `Legacy server ([ruff-lsp](${ruffLspUrl})) has been deprecated. ` +
```
MichaReiser, line 370:
```suggestion
          `The legacy server ([ruff-lsp](${ruffLspUrl})) has been deprecated. ` +
```

src/common/server.ts
```diff
@@ -361,38 +343,39 @@ async function resolveNativeServerSetting(
...
+          `Please consider using the native server instead by setting '${serverId}.nativeServer' to 'on'. `;
+        if (legacyServerSettings.length > 0) {
+          message += `Following settings are not supported with the native server and have been deprecated: ${JSON.stringify(
```
MichaReiser, line 373:
```suggestion
          message += `The following settings are not supported with the native server and have been deprecated: ${JSON.stringify(
```

MichaReiser (approved):

PR #686, opened 2025-02-10
Warn against certain values in `notebook.codeActionsOnSave`

dhruvmanila:
## Summary

This PR adds a check for `notebook.codeActionsOnSave` to do the following:
* For `source.fixAll` and `source.organizeImports`, log at info level
* For `source.fixAll.ruff` and `source.organizeImports.ruff`, log at warn level and provide a warning message to use `notebook.` prefixed code actions instead

## Test Plan

Using the following settings:
```json
{
  "notebook.codeActionsOnSave": {
    "source.fixAll": "explicit",
    "source.organizeImports": "explicit",
    "source.organizeImports.ruff": "explicit",
    "source.fixAll.ruff": "explicit"
  }
}
```

Logs:

```
2025-02-17 20:06:34.796 [info] The following code actions in 'notebook.codeActionsOnSave' could lead to unexpected behavior: ["source.fixAll","source.organizeImports"]. Consider using ["notebook.source.fixAll","notebook.source.organizeImports"] instead. For more information, refer to [this FAQ section](https://docs.astral.sh/ruff/faq/#source-code-actions-in-notebooks).
2025-02-17 20:06:34.796 [warning] The following code actions in 'notebook.codeActionsOnSave' will lead to unexpected behavior: ["source.organizeImports.ruff","source.fixAll.ruff"]. Please use ["notebook.source.organizeImports.ruff","notebook.source.fixAll.ruff"] instead. For more information, refer to [this FAQ section](https://docs.astral.sh/ruff/faq/#source-code-actions-in-notebooks).
```

Notification preview:

Screenshot 2025-02-17 at 8 06 54 PM

### DIFF
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -365,7 +365,7 @@ export function checkNotebookCodeActionsOnSave(serverId: string) {
           genericCodeActions,
         )}. Consider using ${JSON.stringify(
           genericCodeActions.map((action) => `notebook.${action}`),
-        )} instead.`,
+        )} instead. For more information, refer to [this FAQ section](https://docs.astral.sh/ruff/faq/#source-code-actions-in-notebooks).`,
       );
     }
 
@@ -374,7 +374,7 @@ export function checkNotebookCodeActionsOnSave(serverId: string) {
         ruffCodeActions,
       )}. Please use ${JSON.stringify(
         ruffCodeActions.map((action) => `notebook.${action}`),
-      )} instead.`;
+      )} instead. For more information, refer to [this FAQ section](https://docs.astral.sh/ruff/faq/#source-code-actions-in-notebooks).`;
 
       logger.warn(message);
       // Only show a warning if there are Ruff-specific code actions configured.
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -7,6 +7,7 @@ import {
 } from "vscode";
 import { getInterpreterDetails } from "./python";
 import { getConfiguration, getWorkspaceFolders } from "./vscodeapi";
+import { logger } from "./logger";
 
 type ImportStrategy = "fromEnvironment" | "useBundled";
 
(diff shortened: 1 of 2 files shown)


src/common/settings.ts
```diff
@@ -332,6 +333,56 @@ function getPreferredGlobalSetting<T>(
...
+    if (genericCodeActions.length > 0) {
+      // This is at info level because other extensions might be using these code actions but we still want to inform the user.
+      logger.info(
+        `The following code actions in 'notebook.codeActionsOnSave' could lead to unexpected behavior: ${JSON.stringify(
+          genericCodeActions,
+        )}. Consider using ${JSON.stringify(
+          genericCodeActions.map((action) => `notebook.${action}`),
+        )} instead.`,
+      );
+    }
```
dhruvmanila, line 370: I'm a bit unsure if we even want this but a lot of users seems to be using the ones without `.ruff` suffix so at least the logs could help them and us.

I'm also thinking of updating the README to group config to avoid a long list of code snippets which are mostly similar and it's hard to know which config should a user pickup from them.

src/common/settings.ts
```diff
@@ -332,6 +333,56 @@ function getPreferredGlobalSetting<T>(
...
+
+    if (genericCodeActions.length > 0) {
+      // This is at info level because other extensions might be using these code actions but we still want to inform the user.
+      logger.info(
+        `The following code actions in 'notebook.codeActionsOnSave' could lead to unexpected behavior: ${JSON.stringify(
+          genericCodeActions,
+        )}. Consider using ${JSON.stringify(
+          genericCodeActions.map((action) => `notebook.${action}`),
+        )} instead.`,
+      );
```
MichaReiser, line 369: It seems you also added a similar check on the LSP side. Do we only include the warning here because this is a VS code specific fix or is this something we could handle on the LSP side?

Should we include a link to some documentation explaining the why? E.g., we could add a FAQ entry or a section on the website that explains why using the notebook specific actions is preferred.

dhruvmanila: > It seems you also added a similar check on the LSP side. Do we only include the warning here because this is a VS code specific fix or is this something we could handle on the LSP side?

Yes, this is VS Code specific fix as we're looking into `notebook.codeActionsOnSave` which specific to VS Code and can provide suggestion to change it.

> Should we include a link to some documentation explaining why? For example, we could add an FAQ entry or a section on the website explaining why using notebook-specific actions is preferred. It might also be could to link to the issue, but that's something we could do in the FAQ section where we have a bit more space than here.

Thanks, that's a great suggestion. I'll add that first.

MichaReiser (approved):

PR #694, opened 2025-02-18
Show document with server debug info

dhruvmanila:


src/common/commands.ts
```diff
@@ -36,3 +38,86 @@ async function executeCommand(lsClient: LanguageClient, command: string) {
...
+        arguments: [
+          {
+            textDocument: notebookEditor
+              ? { uri: notebookEditor.notebook.uri.toString() }
+              : textEditor
+              ? { uri: textEditor.document.uri.toString() }
+              : undefined,
+          },
+        ],
+      };
```
dhruvmanila, line 83: I updated this in this PR from `main` to check if we're in a notebook context and then use text document. I don't think there's any specific information for a notebook cell so using a notebook for them should be fine.

src/extension.ts
```diff
@@ -188,26 +197,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
-      };
-
-      await lsClient.sendRequest(ExecuteCommandRequest.type, params).then(undefined, async () => {
-        vscode.window.showErrorMessage("Failed to print debug information.");
-      });
-    }),
+    registerCommand(
+      `${serverId}.debugInformation`,
+      createDebugInformationProvider(getClient, serverId, context),
+    ),
```
MichaReiser, line 203: Would it be possible to avoid creating the command if it is the native server?

MichaReiser (approved):
Nice!

src/extension.ts
dhruvmanila: I don't think so. Looking at https://code.visualstudio.com/api/references/contribution-points#contributes.commands, we need to add it to the UI via `package.json` which makes this not possible (https://github.com/astral-sh/ruff-vscode/blob/b4e06981ac3148b3f08c074a759c68f7309cc536/package.json#L434-L438).

PR #700, opened 2025-02-21
Add PPC64 architecture to extension build matrix

### LINKED ISSUE #645: Support PPC64 architecture on Open VSX marketplace
Ruff works flawlessly on the ppc64le architecture. The Ruff VSCode extension also works great under the Codium builds for PPC64 if we manually install the vsix file from x86 and then tell the extension to use the environment version of Ruff instead of bundled. It would be great if you could add ppc64le to the build and deploy target list for the Open VSX marketplace to provide a "native" extension for that platform. I tried to create a fork and make the changes myself but got hopelessly lost trying to figure out exactly what was necessary.

tom-miller1:
## Summary

This enables a build matrix entry for the PPC64le architecture and publication to the OpenVSX registry. I did not include the build target in the VSCode Marketplace registry since Microsoft does not publish an official VSCode build for this architecture. PPC64le is currently only supported by VSCodium.

Resolves #645 

## Test Plan

Ruff binary build is already provided for this architecture, this PR simply adds bundle creation and publishing for this architecture to OpenVSX registry.


tom-miller1:
Closing this as it won't work. The `vsce` process will fail to build the vsix package because it does an explicit target check, and ppc64 is not supported by Microsoft.

MichaReiser:
Hmm. That's a shame. Thanks for trying!

Issue #6, opened 2022-11-09
Support Ruff configuration settings as VS Code extension options


charliermarsh:
E.g., you should be able to set your `line-length` in VS Code directly if desired.

AronDurkinSSS:
@charliermarsh  - how can you _currently_ set line length / disable that error? I've tried editing pyproject.toml and failed. Thanks

charliermarsh:
You can provide a command-line argument directly through the extension settings:

![Screen Shot 2023-01-04 at 4 43 02 PM](https://user-images.githubusercontent.com/1309177/210655558-b7600aa5-ae57-44fc-940f-062ca932b35c.png)

![Screen Shot 2023-01-04 at 4 42 55 PM](https://user-images.githubusercontent.com/1309177/210655576-8a046e53-6e7a-417a-bb6f-59b4ae7676ed.png)

Although better would be to put a `pyproject.toml` in the same path as the files you're linting, and do:

```toml
[tool.ruff]
line-length = 10
```

XuehaiPan:
Hi, have you ever consider allow passing more complex configurations to `ruff` from command line arguments (e.g., a JSON string or similar)? For example, like `clang-format` ([docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#clang-format-style-options)):

> When specifying configuration in the `-style=` option, the same configuration is applied for all input files. The format of the configuration is:
> 
> ```
> -style='{key1: value1, key2: value2, ...}'
> ```

This might need changes in `ruff` upstream. But I think we can discuss it here first.

For C/C++ projects, I'm using [microsoft/vscode-cpptools](https://github.com/microsoft/vscode-cpptools) and I have:

```jsonc
{
    //* C/C++ */
    "C_Cpp.default.configurationProvider": "ms-vscode.cmake-tools",
    "C_Cpp.autocomplete": "Default",
    "C_Cpp.formatting": "clangFormat",
    "C_Cpp.clang_format_sortIncludes": false,
    "C_Cpp.clang_format_style": "file",
    "C_Cpp.clang_format_fallbackStyle": "{ BasedOnStyle: Google, IndentWidth: 4, AccessModifierOffset: -3, ColumnLimit: 100, BinPackArguments: false, BinPackParameters: false }"
}
```

It provides a fallback style if the `.clang-format` file does not exist in the project. I think we can provide a similar thing like:

```jsonc
{
    //* Ruff */
    "ruff.configuration": "file",
[...]

PR #702, opened 2025-02-24
Allow arbitrary config keys in `ruff.configuration`

dhruvmanila:
## Summary

Closes: #6 

This PR adds support for https://github.com/astral-sh/ruff/pull/16296 for the VS Code extension by allowing any arbitrary object in `ruff.configuration`

Additionally, it will provide a warning message to the user if they're using inline configuration but the Ruff version does not support it.

It has one disadvantage that the user will see two popups where the error one is coming from the server because it cannot deserialize the options. We could not show the warning popup if this is too much. I'm worried that it might go unnoticed because the "Show Logs" in the below popup will open the server logs while the message would be in the client logs.

Screenshot 2025-02-25 at 1 24 15 PM

We'll still log it on the console (client log channel):
```
2025-02-25 13:24:10.067 [warning] Inline configuration support was added in Ruff 0.9.8 (current version is 0.9.7). Please update your Ruff version to use this feature.
```

I haven't provided more details in the warning message based on the assumption that if the user is using inline configuration then they're aware of it but it's just that the Ruff version is too old.

## Test Plan

Refer to the test plan in https://github.com/astral-sh/ruff/pull/16296

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
             },
             {
               "type": "object",
-              "markdownDescription": "Set of TOML `<KEY> = <VALUE>` pairs (such as you might find in a `ruff.toml` configuration file) overriding configuration from the filesystem."
+              "markdownDescription": "Inline JSON configuration for Ruff settings (e.g., `{ \"line-length\": 100 }`). *Added in Ruff 0.9.8.*"
             }
           ]
         },
--- a/package.json
+++ b/package.json
@@ -102,9 +102,22 @@
         },
         "ruff.configuration": {
           "default": null,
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -102,9 +102,22 @@
...
+            "object"
+          ],
+          "oneOf": [
+            {
+              "type": "string",
+              "markdownDescription": "Path to a `ruff.toml` or `pyproject.toml` file to use for configuration."
+            },
+            {
+              "type": "object",
+              "markdownDescription": "Set of TOML `<KEY> = <VALUE>` pairs (such as you might find in a `ruff.toml` configuration file) overriding configuration from the filesystem."
```
dhruvmanila, line 118: TODO: need to update this to match the implementation as it's not the same as `--config`

MichaReiser (approved):

PR #703, opened 2025-02-25
Make it easier for Windows users to test Red Knot

InSyncWithFoo:
## Summary

The extension now also recognizes `red_knot.exe` as a Red Knot affix.

## Test Plan

None.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -174,8 +174,10 @@ async function createNativeServer(
   const { path: ruffBinaryPath, version: ruffVersion } = ruffExecutable;
 
   logger.info(`Found Ruff ${versionToString(ruffVersion)} at ${ruffBinaryPath}`);
+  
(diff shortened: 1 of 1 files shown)


src/common/server.ts
```diff
@@ -175,7 +175,7 @@ async function createNativeServer(
 
   logger.info(`Found Ruff ${versionToString(ruffVersion)} at ${ruffBinaryPath}`);
 
-  if (!ruffBinaryPath.endsWith("red_knot")) {
+  if (!ruffBinaryPath.endsWith("red_knot") && !ruffBinaryPath.endsWith("red_knot.exe")) {
```
MichaReiser, line 178: I suggest extracting this into a variable: `let is_red_knot = ....` and then use that variable in the checks instead.

dhruvmanila (approved):
Thanks!

PR #726, opened 2025-03-20
Bump 2025.20.0

dylwil3:


dhruvmanila:
We should use the release script written by @AlexWaygood (https://github.com/astral-sh/ruff-vscode/blob/main/scripts/release.py) to avoid forgetting to update the versions in the README :)

The script basically automates everything and we just need to edit the changelog after running the script.

PR #727, opened 2025-03-21
Fix release regex

dylwil3:
The regex for replacing the version number in the README now detects multiple digit major/minor/patch numbers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Python linter and code formatter, written in Rust. Available on the [Visual Stud
 Ruff can be used to replace Flake8 (plus dozens of plugins), Black, isort, pyupgrade, and more,
 all while executing tens or hundreds of times faster than any individual tool.
 
-The extension ships with `ruff==0.10.0`.
+The extension ships with `ruff==0.11.1`.
 
 ## Highlights
 


MichaReiser (approved):

scripts/release.py
```diff
@@ -143,7 +143,7 @@ def update_readme(latest_ruff: Version) -> None:
     """Ensure the README is up to date with respect to our pinned Ruff version."""
     readme_text = README_PATH.read_text()
     readme_text = re.sub(
-        r"The extension ships with `ruff==\d\.\d\.\d`\.",
+        r"The extension ships with `ruff==\d*\.\d*\.\d*`\.",
```
AlexWaygood, line 146:
```suggestion
        r"The extension ships with `ruff==\d+\.\d+\.\d+`\.",
```

AlexWaygood (approved):
Tyty (https://xkcd.com/1171/ strikes again)

scripts/release.py
dhruvmanila, line 146: There's one more place where we need to update on line 151.

dhruvmanila (approved):
Thank you!

AlexWaygood:
We could possibly also add some assertions to the script:

```py
assert new_readme_text != old_readme_text, "README not updated -- is the release script out of date?"
```

etc.

PR #728, opened 2025-03-21
Release regex followup

dylwil3:


AlexWaygood (approved):
We could also add some assertions here and in some of the script's other functions (https://github.com/astral-sh/ruff-vscode/pull/727#issuecomment-2742959002)

AlexWaygood:
> But then wouldn't we not be able to run the script if we were doing a release/hotfix that wasn't triggered by an update in `ruff`?

Oh, good point. Feel free to merge, then!

PR #740, opened 2025-04-20
filesystemFirst by default

ianzone:
Workspace settings should have higher priority. User editor settings should not trump team workspace settings.


dhruvmanila:
Hi, thank you for opening this PR but we haven't decided yet if we want to change the default here (refer to https://github.com/astral-sh/ruff/issues/13067). You can +1 the issue if you prefer changing the default here. I'll close this PR for now, we can re-open if we decide to change it. Thanks again.

Issue #741, opened 2025-04-23
Native server raises an error "Ruff received invalid settings from the editor" when navigating into installed packages:


MichaReiser:
Native server raises an error "Ruff received invalid settings from the editor" when navigating into installed packages:

```
2025-04-23 11:17:15.804281841  WARN No settings available for file:///usr/local/lib/python3.12/site-packages/cv2/__init__.py - falling back to default settings
2025-04-23 11:17:15.804285127 ERROR Failed to load settings from `configuration`: error looking key 'workspaceFolder' up: environment variable not found
```

This loses the rules, and causes flickering and constant popups as it refreshes frequently.

`ruff-lsp` behaves much better with just a quiet warning.

```
[Warn  - 11:20:42 AM] Skipping standard library file: /usr/local/lib/python3.12/site-packages/cv2/__init__.py
```

_Originally posted by @brad-phobos in https://github.com/astral-sh/ruff/discussions/15991#discussioncomment-12917485_

dhruvmanila:
@brad-phobos Thanks for reporting! Can you share your VS Code settings for Ruff? It seems that somewhere `workspaceFolder` is not being expanded which is something the extension should already do. Or, is it that you're using a different syntax like `$workspaceFolder` instead of `${workspaceFolder}`?

brad-phobos:
```
        "ruff.importStrategy": "fromEnvironment",
        "ruff.configuration": "${workspaceFolder}/../Tools/ruff.toml",
```
I am inside a devcontainer if that matters

dhruvmanila:
> I am inside a devcontainer if that matters

Do you see the same issue outside devcontainer?

Can you also provide debug logs from both the extension and the server? Refer to https://github.com/astral-sh/ruff-vscode#troubleshooting for more info.

dhruvmanila:
Nevermind, I'm able to reproduce this just fine on the VS Code extension.

dhruvmanila:
Ok, somehow I'm seeing the `${workspaceFolder}` in the global client settings which is causing this to fail:
```
2025-04-29 15:48:05.407556000  INFO global settings: ClientSettings {
    configuration: Some(
        String(
            "${workspaceFolder}/formatter/ruff.toml",
        ),
    ),
```

dhruvmanila:
Oh right, the global settings doesn't expand these variables because they don't have a reference i.e., the workspace that these settings belong to.

dhruvmanila:
[...]

PR #746, opened 2025-05-01
Reset `ruff.configuration` if it contains VS Code variables

dhruvmanila:
## Summary

fixes: #741 

## Test Plan

TODO


src/common/settings.ts
```diff
@@ -196,16 +196,36 @@ function getOptionalGlobalValue<T>(config: WorkspaceConfiguration, key: string):
...
+ * resolved. The language server does not support these variables, so they are
+ * filtered out. For example, if `configuration` has either of these variables,
+ * it will be set to `null`.
+ */
 export async function getGlobalSettings(namespace: string): Promise<ISettings> {
   const config = getConfiguration(namespace);
+
+  let configuration = getGlobalValue<string | object | null>(config, "configuration", null);
+  if (typeof configuration === "string" && configuration.search(/\$\{workspaceFolder/) !== -1) {
+    logger.info(
```
MichaReiser, line 215: Should we log this with `warn` or even error?

MichaReiser (approved):
I'm wondering if the extension should fail in that case instead of reverting to a default configuration. Just to be safe and the log message will guide users towards taking the right action. This is, from what I understand, also the closest to the current behavior. 

What was your motivation for returning `none` instead?

dhruvmanila:
> I'm wondering if the extension should fail in that case instead of reverting to a default configuration. Just to be safe and the log message will guide users towards taking the right action. This is, from what I understand, also the closest to the current behavior.

I don't think the extension can fail because it's not a user error. We're passing two settings to the language server - extension settings and global settings. The extension settings are computed for each workspace in the VS Code session while the global settings are computed by just considering the global values. Now, when a user has just specified a single `ruff.configuration` value in the global VS Code settings, the workspace settings will use that and the global settings will also use that. Does that make sense?

src/common/settings.ts
dhruvmanila: I don't think so, it's not a user error but just something to note of. Refer to my other comment.

MichaReiser:
> Now, when a user has just specified a single ruff.configuration value in the global VS Code settings, the workspace settings will use that and the global settings will also use that. Does that make sense?

I don't understand this part. 

I guess the situation I'm concerned about is that we're ignoring a setting explicitly set by the user (in one of their configuration file). Or is this not the case because the `settings.json` is applied boht as workspace and extension settings?

dhruvmanila:
> I guess the situation I'm concerned about is that we're ignoring a setting explicitly set by the user (in one of their configuration file). Or is this not the case because the `settings.json` is applied boht as workspace and extension settings?

We're not ignoring any setting. The language server has access to settings specific to a workspace and the global settings. The global settings is used as a fallback if the workspace doesn't have a settings in the index. When this fallback happens, the server tries to resolve the settings which involves expanding any environment variables which is where the user sees this error. If the settings is used in the context of a workspace, the server will pickup the configuration to the path.

This is a bit confusing because for VS Code, the workspace settings and the default settings (used to get the global settings value) are the same for a VS Code session with the single session unless the user has a workspace specific `settings.json`. Basically, the global `settings.json` values are used for both workspace and global settings. But, if there's `.vscode/settings.json` in the workspace, the workspace settings would use the values from that while the global settings would use the values from the global `settings.json`.

MichaReiser (approved):

PR #765, opened 2025-06-17
Restore version field in `pyproject.toml`

ntBre:
## Summary

This reverts a small part of #750. The `version` field is used by the release script, so I think we should keep it around.

## Test Plan

I added back `2025.22.0` before running the release script, which was then updated to `2025.24.0`. That's why this is stacked on the release branch too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [project]
 name = "ruff-vscode"
+version = "2025.24.0"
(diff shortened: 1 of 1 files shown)


ntBre:
I was just going to merge it after the release PR. I guess that would have made more sense if I hadn't already merged the release PR, oops.

PR #853, opened 2025-10-04
[feature]: implement toggle setting to exclude ruff_cache

### LINKED ISSUE #768: Option to automatically exclude .ruff_cache from file explorer and search
It would improve developer experience if the Ruff VS Code extension offered an option to automatically add `.ruff_cache/` to `files.exclude` and `search.exclude` global settings.

Since `.ruff_cache/` is auto-generated and not meant to be edited or searched, this behavior would align with how VS Code handles other similar folders, e.g., `node_modules/`.

Ideally, the extension would detect the folder and prompt the user to exclude it from the file explorer and search results. This would also demphasize the folder by greying-it-out. Happy to implement if others think this would useful!

11happy:
## Summary

This PR fixes #768 where it detects the folder and prompt the user to exclude it from the file explorer and search results.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -367,9 +367,9 @@
       }
     },
     "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-      "version": "1.1.11",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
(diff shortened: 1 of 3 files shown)


MichaReiser:
Thanks. Could you add a test plan (video) that demonstrates the behavior before/after your change?

PR #1035, opened 2026-05-15
Add support for the Python Environment Extension

### LINKED ISSUE #598: Fallback to using bundled executable if Python interpreter is unavailable
Based on the discussion thread from https://github.com/astral-sh/ruff-vscode/issues/493#issuecomment-2322383690.

Once we switch completely to the native server, it might make sense to use the bundled executable as a fallback and remove the hard requirement of the Python interpreter.

### LINKED ISSUE #1029: Prefer using vscode-python-environments’ interpreter change event
See “Layer 3” comment here: https://github.com/microsoft/vscode-python-environments/issues/1491#issuecomment-4383115486

> When you pick an env via [vscode-python-environments (VPE)]’ tree, we fire `pythonEnvsApi.onDidChangeEnvironment` (per-workspace). Both ruff-vscode and ty-vscode subscribe only to the legacy `api.environments.onDidChangeActiveEnvironmentPath` (global only), so they miss our selections — which is exactly why your `Select Interpreter` workaround "fixes" it (it sets the global env and fires the legacy event). **Pylance already supports both APIs**: it dual-subscribes, prefers ours when our extension is installed, falls back to the legacy event otherwise. We think that's the right model for ruff/ty too.

IDK how they do it (probably buried in some proprietary pylance code) but I can think of one optimal way:

1. check if the `ms-python.vscode-python-envs` extension API can be accessed, if not, load the `ms-python.python` extension
2. subscribe to the loaded extensions’ event
3. subscribe to the extension change event, and when either extension gets disabled, goto 1.

related: #479, https://github.com/astral-sh/ty-vscode/issues/293

MichaReiser:
## Summary

Add dedicated support for the Python Environment Extension

This is mostly a back port from ty-vscode. The main difference is that we need to support ruff-lsp. 

This PR also changes our fallback behavior for the native server. We'll no longer block if the user hasn't selected an interpreter or if it's too old. For those cases, we now fallback to searching PATH and, ultimately, using the bundled executable. 

Closes https://github.com/astral-sh/ruff-vscode/issues/1029
Closes https://github.com/astral-sh/ruff-vscode/issues/598

## Test plan

I tested the interpreter selection with and without the Python environment extension. I verified that the server no longer restarts when using a fixed `ruff.path` (because the interpreter selection is then meaningless). I also tested ruff-lsp.

### DIFF
--- a/src/common/python.ts
+++ b/src/common/python.ts
@@ -174,7 +174,7 @@ class PythonEnvironmentExtension implements EnvironmentProvider {
         const environment = event.new == null ? null : this.toEnvironmentDetails(event.new);
         const previousEnvironment = event.old == null ? null : this.toEnvironmentDetails(event.old);
 
-        if (areEnvironmentsEqual(previousEnvironment, environment)) {
+        if (isDeepStrictEqual(previousEnvironment, environment)) {
           this.#activeEnvironments.remember(event.uri, environment);
           logger.debug(
             `Ignoring Python Environments change event because the active environment is unchanged for '${event.uri ?? "workspace"}'.`,
@@ -268,13 +268,6 @@ function createActiveEnvironmentCache() {
       return !unchanged;
     },
   };
-}
-
-function areEnvironmentsEqual(
-  left: PythonEnvironmentDetails | null,
-  right: PythonEnvironmentDetails | null,
-): boolean {
-  return isDeepStrictEqual(left, right);
 }
 
 export function checkInterpreterVersion(environment: PythonEnvironmentDetails): boolean | null {
--- a/src/common/python.ts
+++ b/src/common/python.ts
@@ -1,79 +1,329 @@
-import { commands, Disposable, Event, EventEmitter } from "vscode";
+import { commands, type Disposable, type Event, EventEmitter, extensions, Uri } from "vscode";
+import {
+  PythonExtension as PythonExtensionApi,
+  type ResolvedEnvironment,
+} from "@vscode/python-extension";
+import type { PythonEnvironment, PythonEnvironmentApi } from "@vscode/python-environments";
 import { logger } from "./logger";
-import { PythonExtension, ResolvedEnvironment, Resource } from "@vscode/python-extension";
 
-export interface IInterpreterDetails {
-  path?: string[];
-  resource?: Resource;
+const onDidChangeActivePythonEnvironmentEvent =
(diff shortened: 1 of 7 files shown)


src/common/python.ts
```diff
@@ -1,79 +1,321 @@
...
+  uri?: Uri;
+};
+
+export const onDidChangeActivePythonEnvironment: Event<OnDidChangeActivePythonEnvironmentEventArgs> =
+  onDidChangeActivePythonEnvironmentEvent.event;
+
+export type PythonCommand = {
+  executable: string;
+  args: string[];
+};
```
MichaReiser, line 24: This is new compared to ty. We need it because ty always calls the executable only without args. Ruff calls it with arguments when starting ruff-lsp

src/common/python.ts
```diff
@@ -1,79 +1,321 @@
...
+
+export interface PythonEnvironmentDetails {
```
MichaReiser, line 26: What's below matches ty

src/common/python.ts
```diff
@@ -1,79 +1,321 @@
...
+  logger.warn(`Python version ${version.major}.${version.minor} is not supported.`);
+  logger.warn(`Selected Python path: '${environment.command?.executable}'`);
+  logger.warn("Supported versions are 3.8 and above.");
   return false;
 }
+
+export async function getDebuggerPath(): Promise<string | undefined> {
+  const api = await getPythonExtensionAPI();
+  return api.debug.getDebuggerPackagePath();
+}
```
MichaReiser, line 291: This is ruff specific and it always requires the python extension API (only needed by ruff-lsp)

src/common/server.ts
```diff
@@ -100,63 +106,138 @@ async function getRuffVersion(executable: string): Promise<VersionInfo> {
  *    which checks the PATH environment variable.
  * 5. If all else fails, return the bundled executable path.
  */
-async function findRuffBinaryPath(settings: ISettings): Promise<string> {
+export type BinaryResolution = {
+  path: string;
+  dependsOnActiveInterpreter: boolean;
+};
+
+export async function resolvePythonEnvironment(
```
MichaReiser, line 114: In ty, this is in `findRuffBinaryPath`. It's a separate function in Ruff because the legacy server only needs the environment and not the binary

charliermarsh (approved):

PR #1061, opened 2026-06-10
Substitute `${env:X}` with an empty string for unset environment variables

notable-equivalent:
## Summary

Change how `${env:X}` blocks are handled so that if the variable `X` is not set, the block is replaced with an empty string, rather than the existing behavior of leaving the entire `${env:X}` block unsubstituted.

This makes this extension more consistent with the behavior of VSCode itself when parsing `${env:X}` blocks (it substitutes an empty string for unset variables). You can observe VSCode's behavior by, for example, creating a task like `{"type":"shell", "label": "echo var", "command": "echo '(${env:SOME_UNSET_VAR})'"}` and seeing that it outputs `()`.

## Test Plan

Set `"ruff.interpreter": [ "${env:SOME_UNSET_VAR}/b" ]` in extension settings. Observed log `Using interpreter: /b` (The previous behavior was `Using interpreter: ${env:SOME_UNSET_VAR}/b`)

Set `"ruff.interpreter": [ "${env:HOME}/b" ]` in extension settings. Observed log `Using interpreter: /home/myname/b` (existing behavior)

### DIFF
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -90,10 +90,7 @@ function resolveVariables(
   value: string | string[],
   workspace?: WorkspaceFolder,
 ): string | string[] | null {
-  const substitutions = new Map<
-    string | RegExp,
-    string | ((substr: string, ...args: any[]) => string)
-  >();
+  const substitutions = new Map<string, string>();
   const home = process.env.HOME || process.env.USERPROFILE;
   if (home) {
     substitutions.set("${userHome}", home);
@@ -105,30 +102,17 @@ function resolveVariables(
   getWorkspaceFolders().forEach((w) => {
     substitutions.set("${workspaceFolder:" + w.name + "}", w.uri.fsPath);
   });
-  // Replace ${env:*} with the environment variable value.
-  substitutions.set(/\${env:([^}]+)}/g, (_substr: string, envName: string): string => {
-    return process.env[envName] ?? "";
-  });
-
   const doSubstitutions = (s: string): string => {
-    let result = s;
     for (const [key, value] of substitutions) {
-      // This if-statement is needed since TypeScript cannot resolve unions against function
-      // overloads, even though both branches do the same thing.
-      if (typeof value === "string") {
-        result = result.replace(key, value);
-      } else {
-        result = result.replace(key, value);
-      }
+      s = s.replace(key, value);
     }
-    return result;
+    return s.replace(/\${env:([^}]+)}/g, (_match, envName) => {
+      const envValue = process.env[envName];
+      return typeof envValue === "string" ? envValue : "";
+    });
   };
 
-  if (typeof value === "string") {
-    return doSubstitutions(value);
-  } else {
-    return value.map(doSubstitutions);
-  }
+  return typeof value === "string" ? doSubstitutions(value) : value.map(doSubstitutions);
 }
 
 export function getInterpreterFromSetting(namespace: string, scope?: ConfigurationScope) {
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -90,7 +90,10 @@ function resolveVariables(
   value: string | string[],
   workspace?: WorkspaceFolder,
 ): string | string[] | null {
-  const substitutions = new Map<string, string>();
+  const substitutions = new Map<
+    string | RegExp,
+    string | ((substr: string, ...args: any[]) => string)
+  >();
   const home = process.env.HOME || process.env.USERPROFILE;
   if (home) {
     substitutions.set("${userHome}", home);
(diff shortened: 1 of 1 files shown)


src/common/settings.ts
```diff
@@ -102,25 +105,29 @@ function resolveVariables(
...
 
-  if (typeof value === "string") {
-    let s = value;
+  const doSubstitutions = (s: string): string => {
+    let result = s;
     for (const [key, value] of substitutions) {
-      s = s.replace(key, value);
+      // This if-statement is needed since TypeScript cannot resolve unions against function
+      // overloads, even though both branches do the same thing.
+      if (typeof value === "string") {
```
notable-equivalent: See https://github.com/microsoft/TypeScript/issues/14107

Also note that this if-statement does end up getting optimized out (by the TypeScript compiler?) in the final output `.js` file

src/common/settings.ts
```diff
@@ -90,7 +90,10 @@ function resolveVariables(
   value: string | string[],
   workspace?: WorkspaceFolder,
 ): string | string[] | null {
-  const substitutions = new Map<string, string>();
+  const substitutions = new Map<
+    string | RegExp,
+    string | ((substr: string, ...args: any[]) => string)
```
notable-equivalent: These are arguments accepted by `String.replace`: A string or a `RegExp` for the search, and a string or function (which is called with the matched string and any capturing groups) for the replacement

MichaReiser:
Thanks for looking into this. Do you have a reference to VS Code's behavior? Can you say more about your use case?

notable-equivalent:
@MichaReiser  VSCode's behavior is not really documented (their [variable reference](https://code.visualstudio.com/docs/reference/variables-reference#_environment-variables) page doesn't say anything about what happens when an env var is unset). However there are a number of ways it can be observed while using VSCode.

#### Observation 1: `tasks.json`

1. Put the following in `.vscode/tasks.json`:
```jsonc
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "echo vars",
            "type": "shell",
            "command": "echo 'UNSET_VAR=(${env:UNSET_VAR}), HOME=(${env:HOME})'",
        }
    ]
}
```

2. In the command palette , select "Tasks: Run Task", then select "echo vars".

3. Observe the terminal output:

```text
Executing task: echo 'UNSET_VAR=(), HOME=(/home/myname)' 

UNSET_VAR=(), HOME=(/home/myname)
```

#### Observation 2: `settings.json`

1. Put the following in `.vscode/settings.json`:

```jsonc
{
    // Replace "linux" with "windows" or "osx" if you're on Windows or Mac
    "terminal.integrated.env.linux": {
        "MY_TEST": "UNSET_VAR=(${env:UNSET_VAR}), HOME=(${env:HOME})",
    }
}
```

2. Open a new terminal

3. Print out the value that VSCode set:
```shell
$ echo "$MY_TEST" 
UNSET_VAR=(), HOME=(/home/nelul)
```

As you can see, in both of these instances VSCode expands `${env:UNSET_VAR}` to the empty string. This behavior is consistent across everywhere that VSCode expands `${env:X}` blocks as far as I can tell.

---

My specific use case is a bit exotic. I wouldn't be suggesting this feature _just_ because it's convenient for my weird use case, I only felt it made sense to open a PR since this makes us more consistent with the existing behavior of VSCode.

<details>

<summary>My specific use case details</summary>

I basically just have an awkward situation caused by the fact that in VSCode settings in `.vscode/settings.json` override settings in `devcontainers.json`.

My company uses some formatting scripts which install the code formatters used by a given repo to a Python venv at `<repo root folder>/.formatting/.formatter-venv`, or to `<repo root folder>/.formatting/.formatter-venv$FORMATTING_VENV_SUFFIX` if the `FORMATTING_VENV_SUFFIX` env variable is set. This is because we need to have different installs of the venv locally and inside dev containers, since they use different Linux versions and can't reuse the same binaries. So locally the folder used is `.formatting/.formatter-venv`, but inside dev containers we have the environment variable `FORMATTING_VENV_SUFFIX` set (say to `-companyname-devcontainer`) so the folder used is `.formatting/.formatter-venv-companyname-devcontainer`.

Now this works fine for our own formatting script, it knows where to look for the binaries based on the env var. However when trying to configure VSCode extensions to use the formatters, it becomes an issue. We would like to just set `"ruff.interpreter": [ "${workspaceFolder}/.formatting/.formatter-venv-companyname-devcontainer/bin/python"]` in the `devcontainer.json` settings, and then if users want to use the formatter outside of the dev container as well they could set `"ruff.interpreter": [ "${workspaceFolder}/.formatting/.formatter-venv/bin/python"]` in their own `.vscode/settings.json`. Unfortunately this does not work because for some inane reason the settings in `.vscode/settings.json` override those in `devcontainers.json`.

Now, we were aware of this issue. For other formatters (e.g. `clang-format`) we solved it by using settings like `"clang-format.executable": "${workspaceFolder}/.formatting/.formatter-venv${env:FORMATTING_VENV_SUFFIX}/bin/clang-format"` inside `.vscode/settings.json`. This works because when the variable is set it expands to `.formatting/.formatter-venv-companyname-devcontainer/bin/clang-format`, and when it is unset it expands to `.formatting/.formatter-venv/bin/clang-format` because the unset var expands to an empty string. So this one setting works both inside and outside the dev container.

However this does not work for ruff (`"ruff.interpreter": [ "${workspaceFolder}/.formatting/.formatter-venv${env:FORMATTING_VENV_SUFFIX}/bin/python"]`) because when the `FORMATTING_VENV_SUFFIX` variable is not set (that being outside of the dev container), the string is left unexpanded as `.formatting/.formatter-venv${env:FORMATTING_VENV_SUFFIX}/bin/python`, instead of what we want which is `.formatting/.formatter-venv/bin/python`.

</details>

notable-equivalent:
@MichaReiser Let me know if you have any more questions ☺️

MichaReiser:
Thank you

PR #1126, opened 2026-07-29
Make both Python extensions optional

### LINKED ISSUE #1114: please remove the required dependency on the new python environments extension
the new python environments extension has caused nothing but problems for myself and many others:

- [its API is not finalized](https://github.com/microsoft/vscode-python-environments#api-reference-proposed), so many extensions don't support it yet
- https://github.com/microsoft/vscode-python-environments/issues/812
- doesn't work in vscodium and other vscode forks that depend on the openVSX marketplace - https://github.com/EclipseFdn/publish-extensions/issues/1002
- they pushed out an A/B test to many users that left the environment settings in a broken state for months - https://github.com/microsoft/vscode-python/issues/25284#issuecomment-3076545360

usually these problems can be avoided by setting `python.useEnvironmentsExtension": false` in `settings.json`, but even just having the extension installed and disabled with this setting breaks the pylint extension - https://github.com/microsoft/vscode-pylint/issues/856

this extension is being forced upon users despite being far from ready. [just look at the ratings and reviews](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-python-envs&ssr=false#review-details) (1.5/5 stars)

it would be great if the dependency on this extension was optional.

ntBre:
Summary
--

This PR closes #1114 by removing the dependency on `ms-python.python` again, without replacing it
with `ms-python.vscode-python-envs`. I documented this in the README, in the preexisting
`Requirements` section, which I moved much closer to the top.

I also updated the logs when the Python extension is unavailable and added a warning notification
too.

Test Plan
--

Manual testing in VS Code, using this command:

<details><summary>Details</summary>
<p>

```console
cd /Users/brw/astral/ruff-vscode

npm run compile

RUFF_DEMO_DIR="$(mktemp -d /private/tmp/ruff-vscode-optional.XXXXXX)"

cp -R src/testFixture "$RUFF_DEMO_DIR/project"

printf '\nimport math\n' >> "$RUFF_DEMO_DIR/project/diagnostics.py"

'/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code' \
  --new-window \
  --skip-welcome \
  --disable-workspace-trust \
  --user-data-dir "$RUFF_DEMO_DIR/data" \
  --extensions-dir "$RUFF_DEMO_DIR/extensions" \
  --extensionDevelopmentPath=/Users/brw/astral/ruff-vscode \
  "$RUFF_DEMO_DIR/project" \
  "$RUFF_DEMO_DIR/project/diagnostics.py"
```

</p>
</details> 

This shows the initial notification window and also that the extension functions without either Python extension installed:

Screenshot 2026-07-29 at 12 22 57

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,9 +36,7 @@
     "Linters",
     "Formatters"
   ],
-  "extensionDependencies": [
-    "ms-python.python"
-  ],
+  "extensionDependencies": [],
   "capabilities": {
     "untrustedWorkspaces": {
       "supported": "limited",
@@ -278,7 +276,7 @@
         },
         "ruff.interpreter": {
           "default": [],
-          "markdownDescription": "Path to a Python interpreter to use to find the `ruff` executable.",
+          "markdownDescription": "Path to a Python interpreter to use to find the `ruff` executable. Requires either the Python Environments or Python extension to be installed.",
           "scope": "resource",
           "items": {
             "type": "string"
--- a/src/common/python.ts
+++ b/src/common/python.ts
@@ -1,11 +1,18 @@
 import { isDeepStrictEqual } from "node:util";
 import { type Disposable, type Event, EventEmitter, extensions, Uri } from "vscode";
 import {
+  PVSC_EXTENSION_ID,
   PythonExtension as PythonExtensionApi,
   type ResolvedEnvironment,
 } from "@vscode/python-extension";
-import type { PythonEnvironment, PythonEnvironmentApi } from "@vscode/python-environments";
+import {
+  EXTENSION_ID as PYTHON_ENVIRONMENTS_EXTENSION_ID,
+  type PythonEnvironment,
+  type PythonEnvironmentApi,
+} from "@vscode/python-environments";
 import { logger } from "./logger";
+
+export { PVSC_EXTENSION_ID, PYTHON_ENVIRONMENTS_EXTENSION_ID };
 
 const onDidChangeActivePythonEnvironmentEvent =
   new EventEmitter<OnDidChangeActivePythonEnvironmentEventArgs>();
@@ -63,7 +70,7 @@ class PythonExtension implements EnvironmentProvider {
   }
 
   static async tryActivate(): Promise<PythonExtension | null> {
-    if (extensions.getExtension("ms-python.python") == null) {
+    if (extensions.getExtension(PVSC_EXTENSION_ID) == null) {
       logger.info("The Python extension is not installed or is disabled.");
       return null;
     }
@@ -138,7 +145,7 @@ class PythonEnvironmentExtension implements EnvironmentProvider {
   }
 
   static async tryActivate(): Promise<PythonEnvironmentExtension | null> {
-    const extension = extensions.getExtension("ms-python.vscode-python-envs");
+    const extension = extensions.getExtension(PYTHON_ENVIRONMENTS_EXTENSION_ID);
 
     if (extension == null) {
       logger.info("The Python Environments extension is not installed or is disabled.");
--- a/src/common/python.ts
+++ b/src/common/python.ts
@@ -63,6 +63,11 @@ class PythonExtension implements EnvironmentProvider {
   }
 
   static async tryActivate(): Promise<PythonExtension | null> {
+    if (extensions.getExtension("ms-python.python") == null) {
+      logger.info("The Python extension is not installed or is disabled.");
+      return null;
+    }
+
     logger.info("Initializing Python extension");
 
     try {
--- a/src/common/python.ts
+++ b/src/common/python.ts
@@ -1,7 +1,7 @@
 import { isDeepStrictEqual } from "node:util";
 import { type Disposable, type Event, EventEmitter, extensions, Uri } from "vscode";
 import {
-  PVSC_EXTENSION_ID,
+  PVSC_EXTENSION_ID as PYTHON_EXTENSION_ID,
   PythonExtension as PythonExtensionApi,
   type ResolvedEnvironment,
 } from "@vscode/python-extension";
@@ -12,7 +12,7 @@ import {
 } from "@vscode/python-environments";
 import { logger } from "./logger";
 
-export { PVSC_EXTENSION_ID, PYTHON_ENVIRONMENTS_EXTENSION_ID };
+export { PYTHON_EXTENSION_ID, PYTHON_ENVIRONMENTS_EXTENSION_ID };
 
 const onDidChangeActivePythonEnvironmentEvent =
   new EventEmitter<OnDidChangeActivePythonEnvironmentEventArgs>();
@@ -70,7 +70,7 @@ class PythonExtension implements EnvironmentProvider {
   }
 
   static async tryActivate(): Promise<PythonExtension | null> {
-    if (extensions.getExtension(PVSC_EXTENSION_ID) == null) {
+    if (extensions.getExtension(PYTHON_EXTENSION_ID) == null) {
       logger.info("The Python extension is not installed or is disabled.");
       return null;
     }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,6 +5,8 @@ import { LazyOutputChannel, logger } from "./common/logger";
 import {
   getEnvironmentProvider,
   onDidChangeActivePythonEnvironment,
+  PVSC_EXTENSION_ID,
+  PYTHON_ENVIRONMENTS_EXTENSION_ID,
   type OnDidChangeActivePythonEnvironmentEventArgs,
 } from "./common/python";
 import { resolveServer, type ServerState, startServer, stopServer } from "./common/server";
@@ -97,9 +99,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         .then((selection) => {
           const extensionId =
             selection === "Python Environments"
-              ? "ms-python.vscode-python-envs"
+              ? PYTHON_ENVIRONMENTS_EXTENSION_ID
               : selection === "Python"
-                ? "ms-python.python"
+                ? PVSC_EXTENSION_ID
                 : undefined;
 
           if (extensionId != null) {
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -278,7 +276,7 @@
         },
         "ruff.interpreter": {
           "default": [],
-          "markdownDescription": "Path to a Python interpreter to use to find the `ruff` executable.",
+          "markdownDescription": "Path to a Python interpreter to use to find the `ruff` executable. Requires either the Python Environments or Python extension to be installed.",
           "scope": "resource",
```
MichaReiser, line 280: Can we create a follow up issue to remove this restriction

src/extension.ts
```diff
@@ -79,6 +79,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 
   const environmentProvider = await getEnvironmentProvider();
 
+  if (environmentProvider == null) {
+    const message =
+      "No Python environment extension is available. Ruff will use a configured, globally " +
+      "installed, or bundled executable. Install the Python Environments or Python " +
+      "extension to enable Python environment detection.";
```
MichaReiser: Any chance we can link to the Readme? The paragraph here is a bit abstract, I'm not sure if users understand it.


```suggestion
      "extension to load a Ruff binary from the activated Python environment.";
```

src/extension.ts
```diff
@@ -79,6 +79,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+
+    logger.info(message);
```
MichaReiser, line 91: What's the reason for logging the message? Is it so that it is always shown. I guess, that makes sense.

src/extension.ts
```diff
@@ -79,6 +79,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+
+    const stateKey = "ruff.pythonEnvironmentRecommendationShown";
+
+    if (!context.globalState.get<boolean>(stateKey)) {
```
MichaReiser, line 95: Nice. I'm surprised that `get` doesn't require an async operation but `update` does...

src/extension.ts
```diff
@@ -79,6 +79,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+      await context.globalState.update(stateKey, true);
+
+      void vscode.window
+        .showInformationMessage(message, "Python Environments", "Python")
+        .then((selection) => {
+          const extensionId =
+            selection === "Python Environments"
+              ? "ms-python.vscode-python-envs"
+              : selection === "Python"
+                ? "ms-python.python"
```
MichaReiser: Let's export those constants from the `python.ts` module.

src/extension.ts
```diff
@@ -79,6 +79,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+                : undefined;
+
+          if (extensionId != null) {
+            void vscode.commands.executeCommand(
+              "workbench.extensions.search",
+              `@id:${extensionId}`,
+            );
```
MichaReiser, line 112: Nice. We may need to add a note that users should restart VS Code after installing, or we should listen to newly installed extensions.

MichaReiser (approved):
Thank you. 

Do we need special handling for ruff-lsp, where, I believe, the extension won't start without any Python environment detection.  That is, when you explicitly configure the legacy server.

Can you test the untrusted workspace mode

src/common/python.ts
```diff
@@ -63,6 +63,11 @@ class PythonExtension implements EnvironmentProvider {
   }
 
   static async tryActivate(): Promise<PythonExtension | null> {
+    if (extensions.getExtension("ms-python.python") == null) {
+      logger.info("The Python extension is not installed or is disabled.");
+      return null;
+    }
+
```
MichaReiser: Codex suggests that you can import the ID

```js
import { PVSC_EXTENSION_ID } from "@vscode/python-extension";
```

package.json
ntBre: https://github.com/astral-sh/ruff-vscode/issues/1128

src/extension.ts
ntBre: Added a note to the popup/log message.

src/extension.ts
ntBre: Yeah, I believe the notification is shown only on the first startup, but the log will be shown each time.

ntBre:
> Do we need special handling for ruff-lsp, where, I believe, the extension won't start without any Python environment detection. That is, when you explicitly configure the legacy server.

Ah yeah, good point. I added a fallback to the native server if no environment extension is available, like the fallback for untrusted workspaces.

> Can you test the untrusted workspace mode

Yep! This should show that the extension still works in untrusted mode (and the updated notification message).

Screenshot 2026-07-30 at 15 53 43

MichaReiser:
> Yep! This should show that the extension still works in untrusted mode (and the updated notification message).


Can you verify that Ruff always picks the bundled binary in an untrusted workspace (just to make sure we didn't regress this. I had no way of testing this before because of the Python extension). Specifically, we should use the bundled extension even when:

* The venv contains a local ruff binary (in fact, we should even skip the venv discovery)
* When a `ruff.binaryPath` is specified
* When a `ruff.interpreterPath` is specified

In short, we should always use the bundled binary

src/common/python.ts
```diff
@@ -1,12 +1,19 @@
 import { isDeepStrictEqual } from "node:util";
 import { type Disposable, type Event, EventEmitter, extensions, Uri } from "vscode";
 import {
+  PVSC_EXTENSION_ID,
```
MichaReiser: Can we also use an alias for this constant, e.g. `PYTHON_EXTENSION_ID`. It feels inconsistent, especially given that we re-export those identifiers.

MichaReiser (approved):

ntBre:
Tested the untrusted workspace support with this command:

<details><summary>Details</summary>
<p>

```console
cd /Users/brw/astral/ruff-vscode

npm run compile

RUFF_DEMO_DIR="$(mktemp -d /private/tmp/ruff-vscode-untrusted.XXXXXX)"

cp -R src/testFixture "$RUFF_DEMO_DIR/project"

printf '\nimport math\n' >> "$RUFF_DEMO_DIR/project/diagnostics.py"

mkdir -p \
  "$RUFF_DEMO_DIR/data/User" \
  "$RUFF_DEMO_DIR/project/.vscode" \
  "$RUFF_DEMO_DIR/project/.venv/bin" \
  "$RUFF_DEMO_DIR/custom-bin"

printf '%s\n' \
  '{' \
  '  "security.workspace.trust.enabled": true,' \
  '  "security.workspace.trust.startupPrompt": "never"' \
  '}' > "$RUFF_DEMO_DIR/data/User/settings.json"

printf '%s\n' \
  '{' \
  '  "ruff.nativeServer": "off",' \
  "  \"ruff.path\": [\"$RUFF_DEMO_DIR/custom-bin/ruff\"]," \
  "  \"ruff.interpreter\": [\"$RUFF_DEMO_DIR/project/.venv/bin/python\"]," \
  '  "ruff.importStrategy": "fromEnvironment",' \
  '  "ruff.trace.server": "verbose"' \
  '}' > "$RUFF_DEMO_DIR/project/.vscode/settings.json"

printf '%s\n' \
  'home = /usr/bin' \
  'include-system-site-packages = false' \
  'version = 3.11.0' \
  > "$RUFF_DEMO_DIR/project/.venv/pyvenv.cfg"

printf '#!/bin/sh\ntouch "%s"\nexit 91\n' \
  "$RUFF_DEMO_DIR/venv-ruff-was-executed" \
  > "$RUFF_DEMO_DIR/project/.venv/bin/ruff"

printf '#!/bin/sh\ntouch "%s"\nexit 92\n' \
  "$RUFF_DEMO_DIR/interpreter-was-executed" \
  > "$RUFF_DEMO_DIR/project/.venv/bin/python"

printf '#!/bin/sh\ntouch "%s"\nexit 93\n' \
  "$RUFF_DEMO_DIR/configured-ruff-was-executed" \
  > "$RUFF_DEMO_DIR/custom-bin/ruff"

chmod +x \
  "$RUFF_DEMO_DIR/project/.venv/bin/ruff" \
  "$RUFF_DEMO_DIR/project/.venv/bin/python" \
  "$RUFF_DEMO_DIR/custom-bin/ruff"

'/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code' \
  --new-window \
  --skip-welcome \
  --user-data-dir "$RUFF_DEMO_DIR/data" \
  --extensions-dir "$RUFF_DEMO_DIR/extensions" \
  --extensionDevelopmentPath=/Users/brw/astral/ruff-vscode \
  "$RUFF_DEMO_DIR/project" \
  "$RUFF_DEMO_DIR/project/diagnostics.py"
```

</p>
</details> 

The logs show us falling back to the bundled executable:

```
2026-07-31 11:43:04.998 [info] Name: Ruff
2026-07-31 11:43:04.998 [info] Module: ruff
2026-07-31 11:43:04.998 [info] The Python Environments extension is not installed or is disabled.
2026-07-31 11:43:04.998 [info] The Python extension is not installed or is disabled.
2026-07-31 11:43:04.998 [info] No Python environment extension is available. Ruff will use a configured, globally installed, or bundled executable. Install the Python Environments or Python extension to load a Ruff binary from the activated Python environment, then reload VS Code. [Learn more](https://github.com/astral-sh/ruff-vscode#requirements).
2026-07-31 11:43:04.998 [warning] Cannot use the legacy server ([ruff-lsp](https://github.com/astral-sh/ruff-lsp)) in an untrusted workspace; switching to the native server using the bundled executable.
2026-07-31 11:43:04.998 [info] Workspace is not trusted, using bundled executable: /Users/brw/astral/ruff-vscode/bundled/libs/bin/ruff
```

In the command above, Codex also shadowed the Ruff and Python executables to create log files if they're invoked, so this shows that they were not:

```console
$ find "$RUFF_DEMO_DIR" -name '*-was-executed'
$
```

I also had Codex repeat this with each setting in isolation, rather than all of them at once, as this example shows.

PR #1127, opened 2026-07-30
Declare support for TOML files

### LINKED ISSUE #27088: errors reported inside `pyproject.toml`/`ruff.toml` are not reported in the language server
### Summary

```toml
# ruff.toml

target-version = "py314"
preview = true

[lint]
extend-select = ["rule-codes-in-selectors"]
ignore = [
  "N818",
]
```

```console
> ruff check .\ruff.toml
rule-codes-in-selectors: [*] Rule code used instead of name in `lint.ignore`
 --> ruff.toml:7:4
  |
5 | extend-select = ["rule-codes-in-selectors"]
6 | ignore = [
7 |   "N818",
  |    ^^^^
8 | ]
  |
help: Replace rule code with `error-suffix-on-exception-name`
  |
6 | ignore = [
  -   "N818",
7 +   "error-suffix-on-exception-name",
8 | ]
  |

Found 1 error.
[*] 1 fixable with the `--fix` option.
```

in vscode however the error is not reported:

Image

(that JSON schema error is a separate issue: https://github.com/SchemaStore/schemastore/issues/6082)

output from Ruff Language Server Trace:

```
[Trace - 4:48:00 PM] Sending notification 'workspace/didChangeWatchedFiles'.
Params: {
    "changes": [
        {
            "uri": "file:///c%3A/Users/user/project/ruff.toml",
            "type": 2
        }
    ]
}


[Trace - 4:48:00 PM] Received request 'workspace/diagnostic/refresh - (1)'.
[Trace - 4:48:00 PM] Sending response 'workspace/diagnostic/refresh - (1)'. Processing request took 1ms
No result returned.
```

### Version

ruff 0.15.22 (0177a7e0d 2026-07-16)

ntBre:
Summary
--

Related to astral-sh/ruff#26862. I copied the Markdown changes from #950.

Closes https://github.com/astral-sh/ruff/issues/27088.

Test Plan
--

Screenshot 2026-07-30 at 12 17 29

### DIFF
--- a/src/common/utilities.ts
+++ b/src/common/utilities.ts
@@ -45,5 +45,6 @@ export function getDocumentSelector(): DocumentSelector {
         { scheme: "vscode-notebook-cell", language: "python" },
         { scheme: "file", language: "markdown" },
         { scheme: "untitled", language: "markdown" },
+        { scheme: "file", pattern: "**/{pyproject.toml,ruff.toml,.ruff.toml}" },
       ];
 }
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ The extension ships with `ruff==0.16.0`.
 ## Usage
 
 Once installed in Visual Studio Code, `ruff` will automatically execute when you open or edit a
-Python or Jupyter Notebook file.
+Python or Jupyter Notebook file, or a Ruff configuration file (`pyproject.toml`, `ruff.toml`, or
+`.ruff.toml`).
 
 If you want to disable Ruff, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension)
 per workspace in Visual Studio Code.
--- a/package.json
+++ b/package.json
@@ -59,7 +59,10 @@
     "onLanguage:markdown",
     "workspaceContains:*.py",
     "workspaceContains:*.ipynb",
-    "workspaceContains:*.md"
+    "workspaceContains:*.md",
+    "workspaceContains:**/pyproject.toml",
+    "workspaceContains:**/ruff.toml",
+    "workspaceContains:**/.ruff.toml"
   ],
   "main": "./dist/extension.js",
   "scripts": {


src/common/utilities.ts
```diff
@@ -45,5 +45,6 @@ export function getDocumentSelector(): DocumentSelector {
         { scheme: "vscode-notebook-cell", language: "python" },
         { scheme: "file", language: "markdown" },
         { scheme: "untitled", language: "markdown" },
+        { scheme: "file", pattern: "**/{pyproject.toml,ruff.toml,.ruff.toml}" },
```
ntBre, line 48: Codex says that this also registers Ruff as a TOML formatter but that fixing this is not simple and should be a follow-up. It also shouldn't prevent users from selecting/configuring an actual TOML formatter, it just shows Ruff as an additional option.

MichaReiser: Can't we change for which documents the LSP registers formatting support

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentFormattingRegistrationOptions

ntBre: Sounds like the answer is yes, but Codex says that fix is back in `ruff_server`, not here. I'll see what that looks like.

MichaReiser: Yes, the fix is in `ruff_server`.

MichaReiser (approved):

src/common/utilities.ts
Wintreist: @ntBre, Hi!
image
After updating to 2026.70.0, ruff started showing that there are errors in the pyproject.toml files.

MichaReiser: Uhh, thanks. I see what's happening here. Any chance you use a Ruff version older or equal than 0.15.0 We unconditionally declare toml support but older ruff versions don't support toml. They assume it's Python code. We need to gate this by ruff version (we should have that available within the extension). @ntBre can you take this on?

Wintreist: Yes, I’m currently on version 0.12.5.
I guess it’s worth updating.

ntBre: Yep, will do! Thanks for the report!

PR #1140, opened 2026-08-05
Remove outdated Markdown formatting preview requirement

AlexWaygood:
## Summary

Remove the outdated claim that formatting Python code blocks in Markdown requires preview mode, along with the unnecessary `ruff.format.preview` setting from the example. Ruff 0.16.0 stabilized Markdown formatting and enables it by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,13 +144,10 @@ And, for Jupyter Notebooks:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Thank you

PR #1147, opened 2026-08-15
Upgrade vscode-languageclient to v10

### LINKED ISSUE #26001: Update to LSP 3.18
Similar to https://github.com/astral-sh/ty/issues/3772 but for Ruff

E.g. notebooks now participate in notebook pull.

MichaReiser:
## Summary

Same as https://github.com/astral-sh/ty-vscode/pull/516 but for Ruff. I summarized the necessary changes in that PR.

Closes https://github.com/astral-sh/ruff/issues/26001

## Test Plan

https://github.com/user-attachments/assets/539b811e-9205-41b0-8eb6-eb13da7b5cbf

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -59,6 +59,38 @@ class ExtensionLogger {
 export const logger = new ExtensionLogger();
 
 /**
+ * Creates an unformatted server output channel that satisfies the language client's log API.
+ *
+ * Ruff already includes timestamps and severity in its logs, so adding VS Code's log formatting
+ * would duplicate both and incorrectly label every stderr message as an error.
+ */
+export function createServerOutputChannel(name: string): vscode.LogOutputChannel {
+  const channel = vscode.window.createOutputChannel(name, "log");
+
+  return {
+    ...channel,
+    get logLevel(): vscode.LogLevel {
+      return vscode.env.logLevel;
+    },
+    onDidChangeLogLevel: vscode.env.onDidChangeLogLevel,
+    trace: logger.trace.bind(logger),
+    debug: logger.debug.bind(logger),
+    info: logger.info.bind(logger),
+    warn: logger.warn.bind(logger),
+    error(error: string | Error, ...args: any[]): void {
+      const message = util.format(error, ...args);
+      // The language client sends both server logs and client errors here. Server logs
+      // include a log level and must keep their original format.
+      if (/\b(trace|debug|info|warn(?:ing)?|error|critical)\b/i.test(message)) {
+        channel.appendLine(message);
+      } else {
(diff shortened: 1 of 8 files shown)


src/common/server.ts
```diff
@@ -293,13 +293,42 @@ async function createNativeServer(
...
+          // clear diagnostics pushed by the server.
+          //
+          // Once the upstream issues are fixed, enable notebook pulls based on a negotiated
+          // client/server capability.
+          return { kind: vsdiag.DocumentDiagnosticReportKind.full, items: [] };
+        }
+
+        return next(document, previousResultId, token);
+      },
+    },
```
MichaReiser, line 331: We can't use the same workaround as ty because Ruff uses a static pull diagnostic registration, which we can't intercept. But, because Ruff doesn't use `interFileDependencies: true`, intercepting it in `provideDiagnostics` is fine without running into https://github.com/microsoft/vscode-languageserver-node/issues/1837.


We'll need to set `interFileDependencies: true` for notebook cells if we want to support pull diagnostics, or VS Code won't pull cells in the same notebook (don't ask me why)

dhruvmanila (approved):

MichaReiser:
I'll wait a few weeks before landing to see how the upgrade plays out for ty

MichaReiser:
I need to carry over the ty-logging fix that I landed a few days later.

dhruvmanila:
We might want to skip this and jump straight to `10.1.1` given that it has https://github.com/microsoft/vscode-languageserver-node/commit/2ffaa4db08c043603f257dd155cf6d204d117b3c

src/common/logger.ts
```diff
@@ -58,6 +58,49 @@ class ExtensionLogger {
...
+  const channel = vscode.window.createOutputChannel(name, "log");
+  const logClient =
+    (level: vscode.LogLevel, label: ClientLogLevel) =>
+    (message: string | Error, ...args: unknown[]): void => {
+      const configuredLevel = vscode.env.logLevel;
+      if (configuredLevel === vscode.LogLevel.Off || configuredLevel > level) {
+        return;
+      }
+      channel.appendLine(
+        `${formatLogTimestamp(new Date())} [${label}] ${util.format(message, ...args)}`,
```
dhruvmanila, line 77: Should we use `.toISOString()` instead?
```js
> new Date().toISOString()
'2026-09-16T10:11:04.043Z'
```

Or a format string if the format is different?

dhruvmanila (approved):
Thank you!

src/common/logger.ts
MichaReiser: This matches ty-vscode's and VS Code's formatting

PR #1155, opened 2026-08-19
fix release script vulnerabilities

woodruffw:


scripts/release.py
```diff
@@ -9,7 +9,7 @@
 """
 
 # /// script
-# requires-python = "==3.8.*"
+# requires-python = ">=3.12"
```
AlexWaygood, line 12: I _think_ this is okay. But I'm not sure. Would like a second opinion from @ntbre!

One of the things this script does is invoke `just lock`, which regenerates `requirements.txt` and `requirements-dev.txt` using Python 3.8. But I don't _think_ there's any reason that this script actually needs to be run using Python 3.8, since `just lock` is invoked as a subprocess from this script anyway.

We also... really probably shouldn't be using `requirements.txt` and `requirements-dev.txt` in this repo in 2026 🙈 this is another one of those very old pieces of infra that's been around for a while and nobody ever updated, I think...

AlexWaygood (approved):

scripts/release.py
ntBre: Oh yeah, this seems fine for the release script! I was initially drafting a reply thinking this was bumping the version for the whole repo, which could affect ruff-lsp.

ntBre (approved):

PR #1163, opened 2026-08-26
Gate TOML support on the Ruff version

ntBre:
## Summary

This PR gates TOML registration behind a Ruff version check to avoid the issue reported in https://github.com/astral-sh/ruff-vscode/pull/1127#discussion_r3862952453, where older versions of Ruff try to lint TOML files as Python code.

The version check uses 0.16.2 because that includes the formatting fix in https://github.com/astral-sh/ruff/pull/27332, in addition to the basic TOML support added to the LSP in https://github.com/astral-sh/ruff/pull/26862 (shipped in 0.16.1) and to the linter in https://github.com/astral-sh/ruff/pull/26772 (shipped in 0.15.22).

TOML files are also excluded when `ruff-lsp` is used.

## Test plan

Manual testing in VS Code:

image

No problem is detected now, but if I use the same executable with the published version of the extension, I reproduce the diagnostic from the comment.

### DIFF
--- a/src/common/status.ts
+++ b/src/common/status.ts
@@ -1,11 +1,16 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { LanguageStatusItem, Disposable, l10n, LanguageStatusSeverity } from "vscode";
+import {
+  LanguageStatusItem,
+  Disposable,
+  l10n,
+  LanguageStatusSeverity,
+  DocumentSelector,
+} from "vscode";
 import { createLanguageStatusItem } from "./vscodeapi";
 import { Command } from "vscode-languageclient";
 import { getDocumentSelector } from "./utilities";
-import { VersionInfo } from "./version";
 
 let _status: LanguageStatusItem | undefined;
 let _serverKind: "native" | "ruff-lsp" | undefined;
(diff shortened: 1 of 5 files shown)


src/common/status.ts
MichaReiser, line 19: Do we also need to pass the version here?

src/common/status.ts
```diff
@@ -23,8 +24,11 @@ export function registerLanguageStatusItem(id: string, name: string, command: st
   };
 }
 
-export function updateServerKind(native: boolean): void {
+export function updateServerKind(native: boolean, ruffVersion?: VersionInfo): void {
   _serverKind = native ? "native" : "ruff-lsp";
+  if (_status) {
+    _status.selector = getDocumentSelector(ruffVersion);
```
MichaReiser: This feels a bit off in `updateServerKind`. It says it updates the server kind, but it now updates the status also. 

It might just be a rename that's necessary here

MichaReiser: I haven't checked. Just wondering if we have access to `client` and if we could use `client.documentSelector` (assuming this exists)

MichaReiser (approved):
Thank you

src/common/status.ts
ntBre: Codex claims that this happens before the version gets resolved and that the initial registration only overs Python and Markdown.

src/common/status.ts
ntBre: Renamed to `updateServerInfo`.

ntBre: Actually, in addressing the `client.documentSelector` suggestion, I ended up restoring this function to its original version and moving the selector update into a separate function called after we obtain a client.

PR #1165, opened 2026-09-02
Normalize workflow filenames, avoid setup-alpine

woodruffw:
This is a preliminary step for a release-gate change similar to the one in ty-vscode. For now it just regularizes our workflow filenames and removes `setup-alpine` because it has an non-hash-pinned transitive action dependency.

Signed-off-by: William Woodruff <william@yossarian.net>


woodruffw:
See https://github.com/astral-sh/ty-vscode/pull/543 and https://github.com/astral-sh/ty-vscode/pull/545 for context.

dhruvmanila (approved):

ntBre (approved):

PR #927, opened 2026-01-09
Remove `ruff-lsp` code

dhruvmanila:
## Summary

Todo:
- [ ] Remove deprecated settings except for `nativeServer` (probably in a follow-up PR?)
- [ ] Is there anything in the CI pipelines that needs to be removed?
- [ ] Test it out!

PR #1170, opened 2026-09-10
Remove ruff-lsp support

### LINKED ISSUE #916: Remove all code and deprecated settings related to ruff-lsp
Now that `ruff-lsp` is completely deprecated and archived, we can remove all code related to ruff-lsp. The deprecated settings should be removed as well. **This change should be tied to a minor Ruff release.**

dhruvmanila:
Remove `ruff-lsp` support so the extension always uses Ruff's native language server. Remove the legacy Python server, settings, bundled dependencies, and release plumbing; retain `ruff.nativeServer` as a deprecated no-op.

This would be released alongside the next major version bump of Ruff and would include a mention of this to redirect any remaining users to the issue tracker for us to understand if there's anything that needs to be done for them to migrate.

Supersedes #927
Resolves #916

## Test plan

- Update diagnostics and server-resolution coverage for the native server
- Cover ignored retired settings and debug information with `nativeServer` disabled
- Remove legacy Python server tests


ntBre (approved):
Nice!

src/common/server.ts
```diff
@@ -305,378 +293,36 @@ async function createNativeServer(
...
+): Promise<ServerResolution> {
+  const resolution = await findRuffBinaryPath(settings, environmentProvider, activeEnvironment);
   return {
-    kind: "legacy",
-    interpreter: interpreter.command,
-    dependsOnActiveInterpreter:
-      (serverSetting.dependsOnActiveInterpreter ?? false) || interpreter.dependsOnActiveInterpreter,
+    executable: {
+      path: resolution.path,
+      version: await getRuffVersion(resolution.path),
```
MichaReiser, line 320: Could we defer resolving the version here to where we check it instead. What would allow us to return `resolution` directly, without this extra wrapper.

src/testFixture/.vscode/settings.json
```diff
@@ -1,8 +1,10 @@
 {
-  // Always use the native server and bundled Ruff binary for the tests.
-  "ruff.nativeServer": "on",
+  // Always use the bundled Ruff binary for the tests.
   "ruff.importStrategy": "useBundled",
 
+  // An old opt-out must not prevent native startup.
+  "ruff.nativeServer": "off",
+
```
MichaReiser, line 7: Should we remove this setting? Or what's the benefit of defaulting to `off`?

package.json
MichaReiser, line 106: Should we update the description to make it clear that they're deprecated and have no effect?

MichaReiser (approved):
Thank you

I think we should show a warning notice (at least on first startup) if a user still uses `off` saying that `ruff-lsp` is no longer supported and that they were migrated to the native ruff server. It's otherwise not clear to users why their setup might break from one day to the other. 

I also wonder if we should keep the other deprecated settings around and show a similar warning, instead of removing them in this minor already.

dhruvmanila:
> I think we should show a warning notice (at least on first startup) if a user still uses `off` saying that `ruff-lsp` is no longer supported and that they were migrated to the native ruff server. It's otherwise not clear to users why their setup might break from one day to the other. 

Yeah, that make sense, will make the change.

> I also wonder if we should keep the other deprecated settings around and show a similar warning, instead of removing them in this minor already.

I don't mind keeping them a bit longer but I'd still suggest to only keep them recognizable but not have them affect anything. So, we could keep them in package.json but remove the code that uses them.

Regardless, I'm not sure what's the benefit of it given that they've been deprecated quite a while ago and some of them were deprecated even before the migration.

src/testFixture/.vscode/settings.json
dhruvmanila: This settings is for the integration  tests that's currently unused. I'll remove them, it doesn't make sense to keep it even for internal use. Thanks for catching this.

MichaReiser:
> I don't mind keeping them a bit longer but I'd still suggest to only keep them recognizable but not have them affect anything. So, we could keep them in package.json but remove the code that uses them.

I think we could still use them to show a warning when using `auto`, the same as we did up to now. Except that we now say, those settings have no effect (or we combine it with the other warning somehow)

package.json
dhruvmanila: There are separate deprecation message fields which VS Code uses (refer to screenshots in the PR description) which gets added to the overall documentation of that field.

dhruvmanila:
> I think we could still use them to show a warning when using `auto`, the same as we did up to now. Except that we now say, those settings have no effect (or we combine it with the other warning somehow)

IIUC, this would be same as today but the message would be different and we'd always use the native server?

src/testFixture/.vscode/settings.json
dhruvmanila: Removed

src/common/server.ts
dhruvmanila: I don't think that is possible, we need to retain the version here because this is also used in `onDidChangeActivePythonEnvironment` to check whether the Ruff version changed or not which would restart the server.

dhruvmanila: We could flatten the structure though

